PR #3, opened 2020-02-13
Clarify that the JSON doc is the doc for matchers

mbernasocchi:
this adresses https://github.com/caddyserver/caddy/issues/3041

### DIFF
--- a/src/docs/markdown/caddyfile/concepts.md
+++ b/src/docs/markdown/caddyfile/concepts.md
@@ -160,7 +160,7 @@ An address must be unique; you cannot specify the same address more than once.
 
 By default, a directive that injects an HTTP handler applies to all requests (unless otherwise documented).
 
-**Request matchers** can be used to classify requests by a given criteria. This concept originates in the [underlying JSON](/docs/json/apps/http/servers/routes/match/) structure, and it's important to know how to use them in the Caddyfile. With matchers, you can specify exactly which requests a directive applies to.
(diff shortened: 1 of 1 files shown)


francislavoie:
That's not exactly true, the Caddyfile matcher docs are directly below the part you edited.

See the block after `The following matcher modules are built-in:`

francislavoie:
FWIW I think Caddyfile matchers should have their own page @mholt, considering how important it is

mbernasocchi:
> That's not exactly true, the Caddyfile matcher docs are directly below the part you edited.
> 
> See the block after `The following matcher modules are built-in:`

Yes I saw that, but that par is super "skinny" where the JSON doc is so much better. Keep in mind that I started using caddy today, so I've got the "perfect noob coming from nginx" perspective :smile:

mholt:
Yeah... and, well, there's a difference between _what a matcher is_ and _which matchers are available_.

I agree we should put which matchers are available on a separate page. In the long run, I also want better inter-linking between the JSON <--> Caddyfile equivalent docs. So I'll close this PR.

mholt:
@mbernasocchi 

> Keep in mind that I started using caddy today, so I've got the "perfect noob coming from nginx" perspective 😄

Great, welcome! Have you seen our WIP [nginx adapter](https://github.com/caddyserver/nginx-adapter)? It lets you run Caddy from an NGINX config (mostly).

mbernasocchi:
not yet, but I'd like to learn caddy to go away from nginx files ;)

PR #4, opened 2020-02-13
add matcher example with capture groups

mbernasocchi:


mholt:
Thanks for the PR. This example seems a bit arbitrary, though, and out of place on this page.

As discussed in the other issue, I think I'll create a new page about the various matchers, and an example like this would be better suited for that page. I'll try to do that soon and update this thread!

mholt:
I've added a full page about request matchers in https://github.com/caddyserver/website/commit/668d145ecc1d136a6e2e14e3fca7186883f7421a (the page is live too)  -- we can continue iterating on that page if desired, so I'll close this PR. Thanks though!

PR #7, opened 2020-02-20
fix the anchor fragment in the index of Caddyfile matchers

mohammed90:
In the words of @francislavoie, the issue this PR solves is:

> TLDR `<h3 id="header-regexp">header_regexp</h3>`, and the links are `<a href="#header_regexp"></a>`


mholt (approved):
Ah, thanks 😁

PR #8, opened 2020-02-21
Add open graph tags and artwork for social media

brunojppb:
## Open Graph Tags
Hi Caddy Team,

A friend just recommended Caddy to me and shared the link. After I tested Caddy a bit and really liked it, I would like contribute with this bit to the website.  
When I got the link via twitter, I couldn't help but notice that there was no meta tags on the page, which I could see when the link didn't bring me any hints about the website.

This pull request aims to add the following features:
- [x] Open Graph tags for social media link sharing
- [x] Artwork that goes together with the Open Graph tags to show off a little about Caddy 

## Open Graph Preview
Below you can see the preview of the Caddy website when shared on social media

### Twitter
Screenshot 2020-02-21 at 13 40 05

### Slack
Screenshot 2020-02-21 at 13 40 40

### Linkedin
Screenshot 2020-02-21 at 13 53 24

## Google Chrome on Android

On Android, we have this great support for theme colors and the meta tag introduced here adds a unique view to the navigation bar. See preview below:

photo5945033473188278566 (1)


mholt (approved):
Wow! This is a fun and useful PR, and excellently documented! Thank you :100:

Feel free to contribute any time!

mholt:
@brunojppb I just pushed a commit that customizes the title tags for each docs page: https://github.com/caddyserver/website/commit/dd3a8ba9cf06b4757e9a6a65ec4034cd6b145781 -- it might be nice to customize the description too. The easiest way to do this might be to just let the bots scape from the body of the page. Do you happen to know if bots will do that if no description is specified? Similarly, do you know if the <title> tag will be used if no meta tags are specified? That would simplify things too.

brunojppb:
@mholt you are definitely right in simplifying that. Although, this only works for search engine crowlers, which means that if we share the website link on social networks, we don't get the correct preview since they specifically target the meta tags.

For instance, I just checked the current version of the website on twitter, and this is how the link previe would be rendered if shared in a tweet:
Screenshot 2020-02-29 at 14 37 20

Slack also uses the `og` tags (falls back to `twitter` tags) to render the link preview. but it also falls back to `<title>` in case we don't specify.

So the downside of not explicitly providing `description` and `title` is that in some social networks the preview won't render properly.

mholt:
@brunojppb Okie, thanks for the handy info. I've updated the homepage to have the meta title tags as well.

PR #10, opened 2020-02-24
Update config import paths for module example

parrotmac:
Looks like the import path is wrong. Sanity checking against an [in-tree module](https://github.com/caddyserver/caddy/blob/v2/modules/filestorage/filestorage.go#L19) seems to confirm this change corrects the import path.

### DIFF
--- a/src/docs/markdown/extending-caddy.md
+++ b/src/docs/markdown/extending-caddy.md
@@ -293,8 +293,8 @@ import (
 	"os"
 
 	"github.com/caddyserver/caddy/v2"
-	"github.com/caddyserver/caddy/v2/config/caddyfile"
-	"github.com/caddyserver/caddy/v2/config/httpcaddyfile"
+	"github.com/caddyserver/caddy/v2/caddyconfig/caddyfile"
+	"github.com/caddyserver/caddy/v2/caddyconfig/httpcaddyfile"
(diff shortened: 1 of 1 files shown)


src/docs/markdown/extending-caddy.md
```diff
@@ -372,4 +372,4 @@ var (
 	_ caddyhttp.MiddlewareHandler = (*Middleware)(nil)
 	_ caddyfile.Unmarshaler       = (*Middleware)(nil)
 )
-```
\ No newline at end of file
+```
```
mholt, line 375: What is the change on this line?

mholt:
Oops, you're right -- thanks!

src/docs/markdown/extending-caddy.md
parrotmac: It looks like the original files doesn't have a trailing newline. (Not sure why GitHub's not rendering with the :no_entry_sign: and arrow it usually shows though)

On `master`:
```
$ xxd src/docs/markdown/extending-caddy.md | tail -1
00003dd0: 286e 696c 290a 290a 6060 60              (nil).).```
```

mholt: Ah, yeah, weird. Usually it shows the arrow. Thanks

mholt (approved):

PR #11, opened 2020-02-26
Add anchor links to headers

francislavoie:
~Uses https://github.com/bryanbraun/anchorjs to add anchor links to headers.~ Now just uses the CSS/font from that lib, but does the rest manually. Simpler.

Screenshot:

image


mholt:
Oh, this is awesome! What a neat surprise 😁 

I like this, but noticed the extra dependency.

Don't we only need something as simple as (untested):

```js
$('article h1[id],h2[id],h3[id],h4[id],h5[id],h6[id]').each(function() {
	$anchor = $('<a href="#'+this.id+'" class="inline-link">link icon here</a>');
	$anchor.prependTo(elem);
});
```

Is it much more complicated than that? Do we need a whole dependency?

francislavoie:
You're absolutely right, rewrote it using a combo of your JS snippet plus the CSS/font from the lib (extracted it from the unminified JS source)

mholt (approved):
Nice! :+1: LGTM. Thank you for the great little improvement!

PR #17, opened 2020-04-04
Clarify the usage of `try_files` in Caddyfile, not supporting matchers

francislavoie:
There's precedence for an explanation like this considering [`php_fastcgi`](https://caddyserver.com/docs/caddyfile/directives/php_fastcgi#expanded-form) has a similar explanation as to when the expanded form is still useful

### DIFF
--- a/src/docs/markdown/caddyfile/directives/try_files.md
+++ b/src/docs/markdown/caddyfile/directives/try_files.md
@@ -29,6 +29,8 @@ The `try_files` directive is basically a shortcut for:
 rewrite @try_files {http.matchers.file.relative}
 ```
 
+Since this directive is a shortcut, it does not allow a matcher in its syntax. If you need a more complex matcher, then use the above expanded form as a basis.
+
 
 ## Examples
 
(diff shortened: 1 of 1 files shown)


src/docs/markdown/caddyfile/directives/try_files.md
```diff
@@ -29,6 +29,8 @@ The `try_files` directive is basically a shortcut for:
 rewrite @try_files {http.matchers.file.relative}
 ```
 
+Since this directive is a shortcut, it does not allow a matcher in its syntax. If you need a more complex matcher, then use the above expanded form as a basis.
```
mholt: Technically, the reason it doesn't allow a matcher isn't because it's a "shortcut" -- for example the php_fastcgi directive has an expanded form and it accepts matchers. It just doesn't accept a matcher, I guess, because it kind of rubs usability wrong in two ways:

- The first argument is usually a path, which often starts with `/`, which looks like a path matcher, which may be unintended/confusing.
- The try_files directive configures a matcher more than a handler (though technically it does both).

It's technically capable of supporting a matcher, but it's more complex and confusing that way I think.

```suggestion
Note that this directive does not accept a matcher token. If you need more complex matching logic, then use the expanded form above as a basis.
```

mholt (approved):
Great, thank you!

PR #31, opened 2020-04-30
docs: Upgrade guide root `*` matcher token, because path starts with `/`

dhoppe:
The syntax has changed with Caddy v2:
* https://caddyserver.com/docs/v2-upgrade#root


francislavoie:
Thanks, well spotted!

PR #34, opened 2020-05-05
Straight rotated aside elements out on hover

mdesantis:
I love the rotate effect on aside elements, but it compromises readability, especially on video elements. Here we straight them out on hover with a good looking ease effect.


mholt:
Thanks for the PR, but, this doesn't seem to work.

What are you having trouble reading? I think it'd be harder to read / more distracting if they moved.

mholt:
Closing due to inactivity

PR #39, opened 2020-05-14
docs: add instructions for Homebrew

zachvalenta:
Figured if Caddy is available in Homebrew might as well add it to the Downloads page.


mholt (approved):
Good idea, thanks!

PR #42, opened 2020-05-16
docs: New common uses page for Caddyfile

mholt:


francislavoie:
Could just add the "caddy" lang labels now, shouldn't break anything

src/docs/markdown/caddyfile/common-uses.md
```diff
@@ -0,0 +1,122 @@
...
+- [Static file server](#static-file-server)
+- [Reverse proxy](#reverse-proxy)
+- [PHP](#php)
+- [Redirect `www.` subdomain](#redirect-www-subdomain)
+- [Trailing slashes](#trailing-slashes)
+
+
+## Static file server
+
+```
```
francislavoie:
```suggestion
```caddy
```

src/docs/markdown/caddyfile/common-uses.md
```diff
@@ -0,0 +1,122 @@
...
+root * /var/www  # optional; default root is current directory
+file_server
+```
+
+
+## Reverse proxy
+
+All requests:
+
+```
```
francislavoie:
```suggestion
```caddy
```

src/docs/markdown/caddyfile/common-uses.md
```diff
@@ -0,0 +1,122 @@
...
+example.com
+
+reverse_proxy localhost:5000
+```
+
+Just requests having a path starting with `/api/`; static files for everything else:
+
+```
```
francislavoie:
```suggestion
```caddy
```

src/docs/markdown/caddyfile/common-uses.md
```diff
@@ -0,0 +1,122 @@
...
+root * /var/www  # optional; default root is current directory
+file_server
+```
+
+
+## PHP
+
+With FastCGI running, something like this works for most modern PHP apps:
+
+```
```
francislavoie:
```suggestion
```caddy
```

src/docs/markdown/caddyfile/common-uses.md
```diff
@@ -0,0 +1,122 @@
...
+example.com
+
+php_fastcgi unix//run/php/php-fpm.sock
+```
+
+If your PHP site relies on static files too, you may need to enable a static file server (but this depends on your PHP app):
+
+```
```
francislavoie:
```suggestion
```caddy
```

src/docs/markdown/caddyfile/common-uses.md
```diff
@@ -0,0 +1,122 @@
...
+php_fastcgi /blog/* localhost:9000
+root * /var/www  # optional; default root is current directory
+file_server
+```
+
+## Redirect `www.` subdomain
+
+To **add** the `www.` subdomain with an HTTP redirect:
+
+```
```
francislavoie:
```suggestion
```caddy
```

src/docs/markdown/caddyfile/common-uses.md
```diff
@@ -0,0 +1,122 @@
...
+
+www.example.com {
+	...
+}
+```
+
+
+To **remove** it:
+
+```
```
francislavoie:
```suggestion
```caddy
```

src/docs/markdown/caddyfile/common-uses.md
```diff
@@ -0,0 +1,122 @@
...
+
+## Trailing slashes
+
+You will not usually need to configure this yourself; the [`file_server` directive](/docs/caddyfile/directives/file_server) will automatically add or remove trailing slashes from requests by way of HTTP redirects, depending on whether the requested resource is a directory or file, respectively.
+
+HTTP redirects are external, but you can internally [`rewrite`](/docs/caddyfile/directives/rewrite) add the slash if you want both URIs to be used for the same resource.
+
+To add or remove a trailing slash:
+
+```
```
francislavoie:
```suggestion
```caddy
```

src/docs/markdown/caddyfile/common-uses.md
```diff
@@ -0,0 +1,122 @@
...
+example.com
+
+rewrite /add     /add/
+rewrite /remove/ /remove
+```
+
+To perform the equivalent change externally (with a redirect), simply replaces `rewrite` with [`redir`](/docs/caddyfile/directives/redir):
+
+```
```
francislavoie:
```suggestion
```caddy
```

src/docs/markdown/caddyfile/common-uses.md
```diff
@@ -0,0 +1,122 @@
+---
+title: Common Caddyfile Uses
+---
+
+# Common Caddyfile Uses
+
+This page presents some complete, minimal Caddyfile configurations for common use cases, which you might find helpful when building your own configs.
+
+These are not drop-in solutions; you will have to customize your domain name, ports/sockets, directory paths, etc. They are intended to illustrate only the basic requirements for the most common use cases.
```
francislavoie: These don't really involve "requirements" IMO, I think this might be better?

```suggestion
These are not drop-in solutions; you will have to customize your domain name, ports/sockets, directory paths, etc. They are intended to illustrate some of the most common configuration patterns.
```

src/docs/markdown/caddyfile/common-uses.md
```diff
@@ -0,0 +1,122 @@
...
+
+reverse_proxy /api/* localhost:5000
+root * /var/www  # optional; default root is current directory
+file_server
+```
+
+
+## PHP
+
+With FastCGI running, something like this works for most modern PHP apps:
```
francislavoie:
```suggestion
With a PHP FastCGI service running, something like this works for most modern PHP apps:
```

src/docs/markdown/caddyfile/common-uses.md
```diff
@@ -0,0 +1,122 @@
...
+
+```
+example.com
+
+php_fastcgi unix//run/php/php-fpm.sock
```
francislavoie: Still need to serve files in the 95% case.

```suggestion
php_fastcgi unix//run/php/php-fpm.sock

root * /var/www  # optional; default root is current directory 
file_server
```

francislavoie: derp the next one says that (but tbh they almost always will)

francislavoie: Actually yeah - we definitely should always note to add file_server here. I can't think of really any situations where it would _only_ be PHP. I guess if it's a JSON API or something... but...

src/docs/markdown/caddyfile/common-uses.md
```diff
@@ -0,0 +1,122 @@
...
+
+HTTP redirects are external, but you can internally [`rewrite`](/docs/caddyfile/directives/rewrite) add the slash if you want both URIs to be used for the same resource.
+
+To add or remove a trailing slash:
+
+```
+example.com
+
+rewrite /add     /add/
+rewrite /remove/ /remove
```
francislavoie: Would it make sense at all to use `uri` here instead?

```suggestion
uri /remove/ strip_suffix /
```

But I understand if the goal was to show the inverse with rewrite only.

The `rewrite` example here is a bit non-optimal I'd say because they involve duplication. But it's minor.

src/docs/markdown/caddyfile/common-uses.md
mholt: I think it's important for the user to know that they are separate things, we definitely don't want them serving static files if they shouldn't be...

src/docs/markdown/caddyfile/common-uses.md
mholt: Good point, maybe -- the semantics of `rewrite` are different from `uri`.

src/docs/markdown/caddyfile/common-uses.md
francislavoie: Okay then we can add a comment `# the below will serve static files` or something. I'd rather avoid people jumping to the forums saying "I'm getting empty 200s", cause we already get plenty of those!

src/docs/markdown/caddyfile/common-uses.md
```diff
@@ -0,0 +1,122 @@
...
+- [Redirect `www.` subdomain](#redirect-www-subdomain)
+- [Trailing slashes](#trailing-slashes)
+
+
+## Static file server
+
+```
+example.com
+
+root * /var/www  # optional; default root is current directory
```
carlwgeorge: I'd leave off the comment here.  Most people run webservers as daemons, not from the command line, so for the "common uses" theme this comment is noise.

src/docs/markdown/caddyfile/common-uses.md
```diff
@@ -0,0 +1,122 @@
...
+
+To **add** the `www.` subdomain with an HTTP redirect:
+
+```
+example.com {
+	redir https://www.example.com{uri}
+}
+
+www.example.com {
+	...
```
carlwgeorge: A `root` directive here might be better here, just to make sure no one gets mixed up thinking that `...` is some magic directive.  Same with a few lines later.

src/docs/markdown/caddyfile/common-uses.md
carlwgeorge: This is confusing, users would want one or the other, not both in the same config.  This will make more sense as two separate config blocks.

src/docs/markdown/caddyfile/common-uses.md
```diff
@@ -0,0 +1,122 @@
...
+To add or remove a trailing slash:
+
+```
+example.com
+
+rewrite /add     /add/
+rewrite /remove/ /remove
+```
+
+To perform the equivalent change externally (with a redirect), simply replaces `rewrite` with [`redir`](/docs/caddyfile/directives/redir):
```
carlwgeorge: Is there a concise way we can describe `rewrite` and `redir` here so that most users don't need to follow the links just for a basic understanding of the difference?

carlwgeorge (changes requested):

src/docs/markdown/caddyfile/common-uses.md
francislavoie: How about: "`redir` asks the browser to change the URL, `rewrite` asks Caddy to change the path internally". Make sense?

carlwgeorge: YES, thank you for putting that into words better than I could have.

src/docs/markdown/caddyfile/common-uses.md
carlwgeorge: I agree, the first example should be the most common pattern (include the `file_server` directive).  In fact I would just go with one example, and have text after it that points out that the `file_server` directive isn't necessary if the user knows they don't need to serve static files.

src/docs/markdown/caddyfile/common-uses.md
mholt: A root directive by itself is not particularly useful though.

I think -- hope (for the sake of humanity) -- that most people in most cultures know that `...` indicates a continuation or filler of some sort.

Maybe the block should just be left empty, if it's really going to be a problem?

francislavoie: I don't feel that strongly either way, but I think `...` is fine. They'll get a syntax error immediately if they try that anyways.

carlwgeorge: How about a `root` and `file_server`, just something minimal that is real?

mholt: I think that starts distracting from what the example is intending to showcase: the redirect and _only_ the redirect.

I'll leave the block empty then.

mholt:
Thanks for all the feedback. Round two!

mholt:
I wonder if the page should be renamed Caddyfile Patterns or something.

src/docs/markdown/caddyfile/common-uses.md
```diff
@@ -111,11 +110,18 @@ rewrite /add     /add/
 rewrite /remove/ /remove
```
francislavoie: I think you forgot the add here

```suggestion
rewrite /add     /add/
rewrite /remove/ /remove
```

mholt: Is that not https://github.com/caddyserver/website/pull/42/files#diff-9f61e9f747cde1dd3e210174ef80eda0R109 ?

francislavoie: The heck, it didn't show up for me 🤷‍♂️

src/docs/markdown/caddyfile/common-uses.md
```diff
@@ -0,0 +1,127 @@
...
+example.com
+
+root * /var/www
+php_fastcgi /blog/* localhost:9000
+file_server
+```
+
+Customize the site root and path matcher accordingly; this example assumes PHP is only in the `/blog/` subdirectory&mdash;all other requests will be served as static files.
+
+The [`php_fastcgi` directive](/docs/caddyfile/directives/php_fastcgi) is actually just a shortcut for [several pieces of configuration](/docs/caddyfile/directives/php_fastcgi#expanded-form).
```
francislavoie:
```suggestion
The [`php_fastcgi` directive](/docs/caddyfile/directives/php_fastcgi) is actually a shortcut for [several pieces of configuration](/docs/caddyfile/directives/php_fastcgi#expanded-form).
```

I wouldn't say it's _just_ a shortcut 😛

src/docs/markdown/caddyfile/common-uses.md
```diff
@@ -0,0 +1,127 @@
...
+
+
+## Redirect `www.` subdomain
+
+To **add** the `www.` subdomain with an HTTP redirect:
+
+```caddy
+example.com {
+	redir https://www.example.com{uri}
```
francislavoie: Do we want to consider making it like this?

```suggestion
	redir https://www.{host}{uri}
```

src/docs/markdown/caddyfile/common-uses.md
```diff
@@ -0,0 +1,127 @@
...
+www.example.com {
+}
+```
+
+
+To **remove** it:
+
+```caddy
+www.example.com {
+	redir https://example.com{uri}
```
francislavoie: Probably too "complex" to show in this page, but maybe:

```suggestion
	redir https://{labels.1}.{labels.0}{uri}
```

carlwgeorge:
> I wonder if the page should be renamed Caddyfile Patterns or something.

I'd rather see this page called "Quick-start" and the other "Quick-starts" pages be renamed to something like "Common Uses".

src/docs/markdown/caddyfile/common-uses.md
mholt: I'd rather keep the examples as simple as possible. Also, this would omit the port (if non-standard).

src/docs/markdown/caddyfile/common-uses.md
mholt: Agreed

mholt:
@carlwgeorge This page is only about the Caddyfile, but the quick-start guides cover caddy as a whole, including its API and command line.

I'll rename this file though and merge in a few minutes.

Issue #44, opened 2020-05-17
Mention file_server from the root directive docs page


dontlaugh:
Migrating from v1 to v2 Caddyfile here. I might be dense, but it took me a while to get caddy to just serve my index.html file. I needed `file_server` as well as `root`. I eventually had success with this configuration:

```
coleman.codes {
    root * /opt/blog/public
    file_server
}
```
You might consider linking to the `file_server` directive from the `root` directive's docs page, or perhaps providing an example that combines both on both pages.

francislavoie:
That's good feedback! Thanks, I'll look into it.

francislavoie:
See #45, is that what you had in mind?

dontlaugh:
Yes! That's going to help folks.

PR #45, opened 2020-05-17
docs: Cross-link the `root` directive with `file_server`/`php_fastcgi`

francislavoie:
As suggested in #44 

This makes it clearer that `file_server` and `php_fastcgi` somewhat depend on `root`, and that `root` should be paired with either of them to be useful.

### DIFF
--- a/src/docs/markdown/caddyfile/directives/root.md
+++ b/src/docs/markdown/caddyfile/directives/root.md
@@ -23,22 +23,29 @@ Note that a matcher token is usually required since the first argument is a path
 
 Set the site root to `/home/user/public_html` for all requests:
 
+(Note that a [wildcard matcher](/docs/caddyfile/matchers#wildcard-matchers) is required here because the first argument is ambiguous with a [path matcher](/docs/caddyfile/matchers#path-matchers).)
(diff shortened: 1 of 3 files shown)


src/docs/markdown/caddyfile/directives/root.md
```diff
@@ -23,22 +23,29 @@ Note that a matcher token is usually required since the first argument is a path
...
 ```
 root * /home/user/public_html
 ```
 
-(A [wildcard matcher](/docs/caddyfile/matchers#wildcard-matchers) is required in this case because the first argument is ambiguous with a [path matcher](/docs/caddyfile/matchers#path-matchers).)
-
 Set the site root to `public_html` (relative to current working directory) for all requests:
 
+(No matcher token is required here because our site root is a relative path, so it does not start with a forward slash and thus is not ambiguous.)
+
```
francislavoie, line 37: FYI I moved these up because I feel it reads better if each example ends with their code block, otherwise it feels like the note has to do with the following example.

mholt (approved):
LGTM. Thanks for kickstarting this.

PR #47, opened 2020-05-19
Clarify that Caddy has no _runtime_ dependencies

carlwgeorge:


mholt (approved):
Well, _technically_ it does depend on an operating system.

And some hardware: CPU, memory, I/O bus to disk and network... whooie

PR #48, opened 2020-05-21
docs: Add many examples for Caddyfile request matchers, move TOC

francislavoie:
Everyone's been requesting more examples in the docs. It's a hard balance, but here's a first attempt.

I also moved the standard matchers TOC up to the top, it always an extra click to get to for no real benefit. I asked @whitestrake and he agreed.

Also shoutouts to @whitestrake for the help with the `not` matcher examples, tricky to get the point across clearly on that one.

Note that some of these examples kinda broke the syntax highlighter... found some more edge cases. I'll try to go through those soon, but I don't think they're really deal-breaking on merging this. I think I might need to make a 3rd variant like `caddy-m` or something for places in the docs where it's only matchers, not in a directive.

### DIFF
--- a/src/docs/markdown/caddyfile/matchers.md
+++ b/src/docs/markdown/caddyfile/matchers.md
@@ -261,7 +261,7 @@ Matches request by the `Host` header field of the request. It is not common to u
 #### Example:
 
 ```caddy-d
-method sub.example.com
+host sub.example.com
 ```
 
 
--- a/src/docs/markdown/caddyfile/matchers.md
+++ b/src/docs/markdown/caddyfile/matchers.md
@@ -14,6 +14,18 @@ title: Request matchers (Caddyfile)
 	- [Path matchers](#path-matchers)
 	- [Named matchers](#named-matchers)
 - [Standard matchers](#standard-matchers)
+	- [expression](#expression)
+	- [file](#file)
+	- [header](#header)
+	- [header_regexp](#header-regexp)
+	- [host](#host)
+	- [method](#method)
+	- [not](#not)
+	- [path](#path)
+	- [path_regexp](#path-regexp)
+	- [protocol](#protocol)
+	- [query](#query)
+	- [remote_ip](#remote-ip)
 
 
 ## Syntax
@@ -121,19 +133,6 @@ For most matchers that accept multiple values, those values are OR'ed; i.e. one 
 
 Full matcher documentation can be found [in each respective matcher module's docs](/docs/json/apps/http/servers/routes/match/).
 
-- [expression](#expression)
-- [file](#file)
-- [header](#header)
-- [header_regexp](#header-regexp)
-- [host](#host)
-- [method](#method)
-- [not](#not)
-- [path](#path)
-- [path_regexp](#path-regexp)
-- [protocol](#protocol)
-- [query](#query)
-- [remote_ip](#remote-ip)
-
 
 
 ### expression
@@ -148,11 +147,19 @@ By any [CEL (Common Expression Language)](https://github.com/google/cel-spec) ex
 
 As a special case, Caddy [placeholders](/docs/conventions#placeholders) (or [Caddyfile shorthands](/docs/caddyfile/concepts#placeholders)) may be used in these CEL expressions, as they are preprocessed and converted to regular CEL function calls before being interpreted by the CEL environment.
 
-Examples:
+#### Examples:
 
 ```caddy-d
+# Match requests whose methods start with "P", e.g. PUT or POST.
 expression {method}.startsWith("P")
 ```
+
+```caddy-d
+# Match requests where handler returned error status code 404,
+# would be used in conjunction with the handle_errors directive.
+expression {http.error.status_code} == 404
+```
+
 
 
 ### file
@@ -179,6 +186,24 @@ By files.
 
 An empty `file` matcher will see if the requested file (verbatim from the URI, relative to the [site root](/docs/caddyfile/directives/root)) exists.
 
+Since rewriting based on the existence of a file on disk is so common, there is also a [`try_files` directive](/docs/caddyfile/directives/try_files) which is a shortcut of the `file` matcher and a [`rewrite` handler](/docs/caddyfile/directives/rewrite).
+
+#### Examples:
+
+```caddy-d
+# Match requests where the path is a file that exists.
+file
+```
+
+```caddy-d
+# Match requests where the path followed by .html is a file that
+# exists, or if not, where the path is a file that exists.
+file {
+	try_files {path}.html {path} 
+}
+```
+
+
 
 ### header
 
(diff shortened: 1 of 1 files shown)


mholt (approved):
Thanks! This will be great.

Is there any reason we need to use Caddyfile comments to explain the examples though? On most other docs pages we just use regular text to explain the examples. I don't feel strongly either way, just wondering.

francislavoie:
Not really, just felt it thought it might read better? I'm not really attached either way but felt like trying it. If you want me to switch it I can do that too.

mholt:
Caddyfile comments don't get wrapped because they're in code blocks, making it harder to read on mobile (yeah I know, why are you reading web server docs on your phone), so that's one thing to consider. Another likely side-effect is that people will copy-paste the comments into their own configs and leave them there, even if they change the non-commented lines (thus leading to misleading comments).

User errors, and minor edge cases perhaps, but I think generally I prefer to keep explanation outside of code blocks. I'll let you decide!

francislavoie:
wrapping on mobile is a good point. I'll fiddle with the formatting this evening :+1:

francislavoie:
I didn't find it easy visually to tell where the docs for one matcher began and the other ended, so I added `---` (`<hr>`) between each.

src/docs/markdown/caddyfile/matchers.md
```diff
@@ -213,16 +258,38 @@ host <hosts...>
 
 Matches request by the `Host` header field of the request. It is not common to use this in the Caddyfile, since most site blocks already indicate hosts in the address of the site. This matcher is mostly used in site blocks that don't define specific hostnames.
 
+#### Example:
 
+```caddy-d
+method sub.example.com
```
mholt:
```suggestion
host sub.example.com
```

src/docs/markdown/caddyfile/matchers.md
```diff
@@ -239,7 +306,40 @@ not {
...
+
+```caddy-d
+not path /css/* /js/*
+```
+
+Match requests WITH NEITHER:
+- an `/api/` path prefix, NOR
+- the `POST` request method
+
+i.e. can have none of these to match
```
mholt:
```suggestion
i.e. cannot match either of these:
```

src/docs/markdown/caddyfile/matchers.md
```diff
@@ -239,7 +306,40 @@ not {
...
+```caddy-d
+not path /api/*
+not method POST
+```
+
+Match requests WITHOUT BOTH:
+- an `/api/` path prefix, AND
+- the `POST` request method
+
+i.e. can have zero or one of these to match
```
mholt:
```suggestion
i.e. can match neither or either, but not both:
```

mholt (changes requested):
Thanks Francis, I like the examples a lot!

Just a few wording nits... not sure if they're much better but they made more sense to me.

src/docs/markdown/caddyfile/matchers.md
francislavoie: Good catch, whoops 😄

src/docs/markdown/caddyfile/matchers.md
francislavoie: I don't think I agree - the entire thing is the matcher, not those individually. The implication is that this is in a named matcher, and each of those are rules. Agree on the `:` though.

src/docs/markdown/caddyfile/matchers.md
francislavoie: I think `i.e. can have neither or either of these to match:` would work.

src/docs/markdown/caddyfile/matchers.md
mholt: Well they are also matchers, it is matchers within matchers. I don't think people will care too much, do you think they won't know what we mean when we say "both"?

src/docs/markdown/caddyfile/matchers.md
mholt: Or what about "must not match both"

francislavoie: I think it's important to specify that it can match if zero or one of them to match. Otherwise, it's ambiguous what happens in that case if we don't say so.

src/docs/markdown/caddyfile/matchers.md
francislavoie: The entire point of the example is to explain the behaviour of multiple matchers with `not`s in the context of a named matcher. IMO we should be consistent here when using the word `match` to always refer to the same thing (the named matcher) instead of either referring to the individual matchers or the named matcher.

mholt: Uh, well how about we call them "inner" matchers then? I know it's confusing, but these are advanced cases, after all. At some point, some thinking is required. :)

I just think "can have none of these to match" is both awkward English and more confusing than "cannot match either of these" -- I assume most people will be able to know what "these" refers to?

src/docs/markdown/caddyfile/matchers.md
mholt: Okay, "can have neither or either of these to match" sounds good to me then.

src/docs/markdown/caddyfile/matchers.md
whitestrake: I was the one who suggested "can have none of these to match". It's not technically incorrect, but you're right, it sounds awkward because that kind of phrasing is not common at all in conversation.

Perhaps "must have none of these to match". The intent was to follow from the wording from earlier: "can have zero or one of these to match", which could also be changed to "must".

francislavoie: I was just about to suggest that @whitestrake 👍

src/docs/markdown/caddyfile/matchers.md
```diff
@@ -329,7 +329,7 @@ Match requests WITHOUT BOTH:
 - an `/api/` path prefix, AND
 - the `POST` request method
 
-i.e. can have zero or one of these to match
+i.e. can have neither or either of these to match:
```
whitestrake: Would it be overcritical to request we change this to "must"?

I do very much like calling back to familiar wording and would like to keep the two as similar as possible so the difference is _exceptionally_ clear in comparison; "none" vs. "neither or either", rather than "must have none" vs "can have neither or either".

francislavoie: I feel like `can` and `or` go hand-in-hand. I feel like `can` is the better word if there's multiple options, `must` being better if there's only one option

whitestrake: Conversationally, `must` is exhaustive while `can` is permissive and open ended but not _strictly_ prohibitive beyond the stated options. It's ambiguous. If we're writing technically, "can" functions for both cases and we can ignore the conversational awkwardness because it's a technical statement. If we're writing conversationally, I think "must" is important in both cases.

francislavoie: Fair enough.

mholt (approved):
Thanks, both!

PR #56, opened 2020-06-08
Fix wrong syntax for srv+ prefixed addresses

danlsgiga:
SRV DNS records provide the host + port via DNS resolution, so specifying a port for upstream addresses with the `srv+` prefix is a non pattern, IMO.

I've also added the `srv+https` example


francislavoie (approved):

mholt (approved):
Thank you very much!

PR #63, opened 2020-07-22
Emphasize the lack of implicit `file_server`.

lsb:
I had thought I read the _Primary Changes_ section, but I know I read the `root` section top to bottom, and I think this would have saved me time: I got no error message when I didn't have a `file_server` directive, I just got a 0-byte HTTP response, and it really confused me.

### DIFF
--- a/src/docs/markdown/v2-upgrade.md
+++ b/src/docs/markdown/v2-upgrade.md
@@ -325,6 +325,8 @@ You may find the new [`route`](/docs/caddyfile/directives/route) and [`handle`](
 
 Because it accepts a matcher in v2, this means you can also change the site root depending on the request.
 
+Please note than any implicit `file_server` directives that v1 inferred are now [required](#primary-changes).
+
 
 ### status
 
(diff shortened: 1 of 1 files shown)


src/docs/markdown/v2-upgrade.md
```diff
@@ -325,6 +325,8 @@ You may find the new [`route`](/docs/caddyfile/directives/route) and [`handle`](
 
 Because it accepts a matcher in v2, this means you can also change the site root depending on the request.
 
+Please note than any implicit `file_server` directives that v1 inferred are now [required](#primary-changes).
```
francislavoie: I'd word it more like this I think, to roughly match the Primary Changes section:

```suggestion
Remember to add a [`file_server` directive](https://caddyserver.com/docs/caddyfile/directives/file_server) if serving static files, since Caddy 2 does not assume this by default, whereas in v1 always had it enabled.
```

Thanks!

lsb: Thanks!

lsb:
(I don't think I'd seen an error in the logs, and I wish I had seen one, but whether the behavior should cause a warning is a different issue)

lsb:
Hi, just wondering your thoughts, thanks

mholt (approved):
LGTM, thank you!

PR #67, opened 2020-07-25
Fix grammar

jergason:
"Comprise" means "to contain" not "to make up", so "comprised of" means "to contain of", which doesn't really make sense.

https://prowritingaid.com/grammar/1000078/-composed-of-vs-comprised-of---Which-is-correct

### DIFF
--- a/src/docs/markdown/caddyfile/matchers.md
+++ b/src/docs/markdown/caddyfile/matchers.md
@@ -98,7 +98,7 @@ Path matcher tokens must start with a forward slash `/`.
 
 ### Named matchers
 
-All matchers that are not composed of a single path matcher or a wildcard matcher must be named matchers. This is a matcher that is defined outside of any particular directive, and can be reused.
+All matchers that are not path or wildcard matchers must be named matchers. This is a matcher that is defined outside of any particular directive, and can be reused.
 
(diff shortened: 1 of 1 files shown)


mholt:
Ha, the grammar police caught me. 👮‍♂️  Nice to see you around these parts, Jamison!

TIL. And I am fine with this change.

Actually what do you think of "do not consist of"? Might read a little more naturally.

src/docs/markdown/caddyfile/matchers.md
```diff
@@ -98,7 +98,7 @@ Path matcher tokens must start with a forward slash `/`.
 
 ### Named matchers
 
-All matchers that are not comprised of a single path matcher or a wildcard matcher must be named matchers. This is a matcher that is defined outside of any particular directive, and can be reused.
+All matchers that are not composed of a single path matcher or a wildcard matcher must be named matchers. This is a matcher that is defined outside of any particular directive, and can be reused.
```
mholt:
```suggestion
All matchers that are not path or wildcard matchers must be named matchers. This is a matcher that is defined outside of any particular directive, and can be reused.
```

mholt:
I think I found a wording that is even simpler. :) Thanks for the suggestion and improvement!

PR #71, opened 2020-08-04
docs: Minor updates to `php_fastcgi` docs

francislavoie:
I noticed while dealing with #3630 that `index` doesn't actually support multiple files. Also updated a bit of the wording to clarify some points.


francislavoie:
Wait a sec on this until caddyserver/caddy#3632 is dealt with, might revert that one change to the expanded form section.

PR #72, opened 2020-08-05
Fix doc typo

ajschmidt8:
This PR fixes a typo in the docs. The rest of this tutorial uses port `2016`. It looks like the `2019` port that was here is a typo.


francislavoie:
Nope, that's not a typo!

Port 2016 is your actual configured server, whereas port 2019 is the [Caddy config API](https://caddyserver.com/docs/api).

That said, @mholt I think it would be clearer if we used obviously different port numbers like `8001` and `8002` etc instead of `2015` and `2016` in the guides.

mholt:
I'm fine with that. (But a lot of other server demos use those ports; try to find something a little more unique.)

PR #73, opened 2020-08-05
docs: Add link to the `file_server browse` default template

francislavoie:


mholt (approved):
That's probably a good idea. Thanks!

PR #74, opened 2020-08-05
docs: Better clarify the `auto_https` options and Caddyfile config

polarathene:
My attempt to improve on the docs as a new Caddy user who got confused a bit with this feature, originally [raised an issue about it thinking I had run into a bug](https://github.com/caddyserver/caddy/issues/3635), but was due to Caddy defaulting to HTTPS with HTTP redirect implicitly, but only making a single port implicitly accessible (I assumed `disable_redirects` would keep HTTPS available but also HTTP just without the redirect).

Happy to iterate on feedback, I understand if it might be a bit too verbose or repetitive at present. I've also tried to expose a bit more discovery to related docs, as those who approach learning / experimentation the way I do may not have come across that info yet.

---

Not sure where [these JSON docs](https://caddyserver.com/docs/json/apps/http/servers/automatic_https/) are located, I take it they're not markdown? I was going to backlink to the main doc page on Automatic HTTPS. Not sure if JSON examples are necessary (Docs seem to prefer caddy snippets usually?).

I didn't see any contribution docs for the repo or style guide / linter, let me know if I need to make any changes regarding that. I think codefences are using tabs for indentation? I've used 2 spaces, but can change that if tabs are convention here.

### DIFF
--- a/src/docs/markdown/automatic-https.md
+++ b/src/docs/markdown/automatic-https.md
@@ -23,7 +23,7 @@ Here's a 28-second video showing how it works:
 - Caddy serves IP addresses and local/internal hostnames over HTTPS with locally-trusted certificates. Examples: `localhost`, `127.0.0.1`.
 - Caddy serves public DNS names over HTTPS with certificates from [Let's Encrypt](https://letsencrypt.org). Examples: `example.com`, `sub.example.com`, `*.example.com`.
 
-Caddy keeps all certificates renewed, and redirects HTTP (default port 80) to HTTPS (default port 443) automatically (provided [Activation](/docs/automatic-https#activation) is successful).
+Caddy keeps all certificates renewed, and redirects HTTP (default port 80) to HTTPS (default port 443) automatically, provided that [activation](/docs/automatic-https#activation) is successful.
 
 **For local HTTPS:**
 
@@ -76,7 +76,7 @@ Automatic HTTPS never overrides explicit configuration.
 
 You can [customize or disable automatic HTTPS](/docs/json/apps/http/servers/automatic_https/) if necessary.
 
-<aside class="tip">Disabling redirects keeps the HTTPS port enabled as the default port implicitly assigned to an address. The HTTP port was only binded for redirects, to support connecting to both ports, you must [explicitly listen on both for an address](/docs/automatic-https#examples).</aside>
+<aside class="tip">Disabling redirects keeps the HTTPS port enabled as the default port implicitly assigned to an address. The HTTP port is only bound for redirects and for the ACME HTTP challenge. If you need to serve over HTTP, you must explicitly configure Caddy to do it. [See the examples](/docs/automatic-https#examples).</aside>
 
 
 ## Hostname requirements
@@ -227,11 +227,11 @@ To get a wildcard from Let's Encrypt, you simply need to enable the [DNS challen
 
 ## Examples
 
-Caddy implicitly uses the HTTPS port (default 443) for your [server addresses](/docs/conventions#network-addresses) that don't assign an explicit port (which would disable automatic HTTPS). 
+Caddy implicitly uses the HTTPS port (default 443) for your [server addresses](/docs/conventions#network-addresses) that don't specify a port explicitly (which would disable automatic HTTPS). 
 
-The global setting `auto_https` has two values:
-- `disable_redirects` adds an implicit HTTP port redirect.
-- `off` disables automatic HTTPS, default implicit port changes to HTTP for all server addresses.
+Automatic HTTPS can be configured via the Caddyfile with the [`auto_https` global option](/docs/caddyfile/options), or via [per-server JSON configuration](/docs/json/apps/http/servers/automatic_https/). The Caddyfile option can be set to either of the following:
+- `disable_redirects` which disables the implicit HTTP->HTTPS redirect.
+- `off` which disables automatic HTTPS altogether, including the HTTP->HTTPS redirect and automatic enabling of TLS for sites that meet the requirements.
 
 | auto_https        | HTTP        | HTTPS       |
 |-------------------|-------------|-------------|
@@ -245,27 +245,27 @@ The global setting `auto_https` has two values:
 
 ### Disabling automatic HTTPS
 
-For local development environments, you can prevent serving via HTTPS by providing an explicit port assignment(disable per server address) or using the global `auto_https off` setting which will change Caddy's implicit port to be the HTTP port (default 80) globally.
+For local development environments, you can prevent serving via HTTPS by either specifying `http://` or providing a non-HTTPS port to disable per site, or by adding `auto_https off` to the Caddyfile global options which will change Caddy's implicit port to be the HTTP port (default 80).
 
 ```caddy
 {
-  auto_https off
+	auto_https off
 }
 
 # Have caddy implicitly use the HTTP port
 localhost {
-  root * /usr/share/caddy
+	root * /usr/share/caddy
 
-  file_server
+	file_server
 }
 ```
 
 ```caddy
 # Alternatively, provide an explicit port
 localhost:9000 {
-  root * /usr/share/caddy
+	root * /usr/share/caddy
 
-  file_server
+	file_server
 }
 ```
 
@@ -278,27 +278,27 @@ To do so, you can [map several addresses to a site block as a list](/docs/caddyf
 
 ```caddy
 {
-  auto_https disable_redirects
+	auto_https disable_redirects
 }
 
 # Uses the HTTP and HTTPS by protocol
 # These are configurable as global settings
 http://localhost, https://localhost {
-  root * /usr/share/caddy
+	root * /usr/share/caddy
 
-  file_server
+	file_server
 }
 ```
 
 ```caddy
 {
-  auto_https disable_redirects
+	auto_https disable_redirects
 }
 
 # Alternatively provide explicit ports
 localhost:80, localhost:443 {
-  root * /usr/share/caddy
+	root * /usr/share/caddy
 
-  file_server
+	file_server
 }
 ```
--- a/src/docs/markdown/automatic-https.md
+++ b/src/docs/markdown/automatic-https.md
@@ -230,6 +230,7 @@ To get a wildcard from Let's Encrypt, you simply need to enable the [DNS challen
 Caddy implicitly uses the HTTPS port (default 443) for your [site addresses](/docs/conventions#network-addresses) that don't specify a port explicitly (which would disable automatic HTTPS). 
 
 Automatic HTTPS can be configured via the Caddyfile with the [`auto_https` global option](/docs/caddyfile/options), or via [per-server JSON configuration](/docs/json/apps/http/servers/automatic_https/). The Caddyfile option can be set to either of the following:
+
 - `disable_redirects` which disables the implicit HTTP->HTTPS redirect.
 - `off` which disables automatic HTTPS altogether, including the HTTP->HTTPS redirect and automatic enabling of TLS for sites that meet the requirements.
 
@@ -276,7 +277,7 @@ localhost:9000 {
 
 If you wish to serve content through both HTTP and HTTPS, without HTTP->HTTPS redirects, your site address must explicitly declare the intent to listen on both ports. To do so, [specify multiple site labels](/docs/caddyfile/concepts#addresses), separated by a comma or whitespace.
 
-HTTP and HTTPS by protocol, default ports can be configured via the Caddyfile [`http_port` and `https_port` global options](/docs/caddyfile/options):
+One way to do this is by specifying the protocol scheme, which infers the standard ports (the respective ports can be customized using [`http_port` and `https_port` global options](/docs/caddyfile/options)):
 
 ```caddy
 {
--- a/src/docs/markdown/automatic-https.md
+++ b/src/docs/markdown/automatic-https.md
@@ -227,7 +227,7 @@ To get a wildcard from Let's Encrypt, you simply need to enable the [DNS challen
 
 ## Examples
 
-Caddy implicitly uses the HTTPS port (default 443) for your [server addresses](/docs/conventions#network-addresses) that don't specify a port explicitly (which would disable automatic HTTPS). 
+Caddy implicitly uses the HTTPS port (default 443) for your [site addresses](/docs/conventions#network-addresses) that don't specify a port explicitly (which would disable automatic HTTPS). 
 
 Automatic HTTPS can be configured via the Caddyfile with the [`auto_https` global option](/docs/caddyfile/options), or via [per-server JSON configuration](/docs/json/apps/http/servers/automatic_https/). The Caddyfile option can be set to either of the following:
 - `disable_redirects` which disables the implicit HTTP->HTTPS redirect.
@@ -247,12 +247,13 @@ Automatic HTTPS can be configured via the Caddyfile with the [`auto_https` globa
 
 For local development environments, you can prevent serving via HTTPS by either specifying `http://` or providing a non-HTTPS port to disable per site, or by adding `auto_https off` to the Caddyfile global options which will change Caddy's implicit port to be the HTTP port (default 80).
 
+Have caddy implicitly use the HTTP port:
+
 ```caddy
 {
 	auto_https off
 }
 
-# Have caddy implicitly use the HTTP port
 localhost {
 	root * /usr/share/caddy
 
@@ -260,8 +261,9 @@ localhost {
 }
 ```
 
+Alternatively, provide an explicit port:
+
 ```caddy
-# Alternatively, provide an explicit port
 localhost:9000 {
 	root * /usr/share/caddy
 
@@ -272,17 +274,15 @@ localhost:9000 {
 
 ### HTTP and HTTPS without redirect
 
-If you want to serve content through both HTTP and HTTPS ports without HTTP redirects, your server address cannot rely on a single implicit port and you must explicitly declare the intent to listen from both ports.
+If you wish to serve content through both HTTP and HTTPS, without HTTP->HTTPS redirects, your site address must explicitly declare the intent to listen on both ports. To do so, [specify multiple site labels](/docs/caddyfile/concepts#addresses), separated by a comma or whitespace.
 
-To do so, you can [map several addresses to a site block as a list](/docs/caddyfile/concepts#addresses) separated with `,`:
+HTTP and HTTPS by protocol, default ports can be configured via the Caddyfile [`http_port` and `https_port` global options](/docs/caddyfile/options):
 
 ```caddy
 {
 	auto_https disable_redirects
 }
 
-# Uses the HTTP and HTTPS by protocol
-# These are configurable as global settings
 http://localhost, https://localhost {
 	root * /usr/share/caddy
 
@@ -290,12 +290,13 @@ http://localhost, https://localhost {
 }
 ```
 
+Alternatively specify the ports explicitly:
+
 ```caddy
 {
 	auto_https disable_redirects
 }
 
-# Alternatively provide explicit ports
 localhost:80, localhost:443 {
 	root * /usr/share/caddy
 
--- a/src/docs/markdown/automatic-https.md
+++ b/src/docs/markdown/automatic-https.md
@@ -23,7 +23,7 @@ Here's a 28-second video showing how it works:
 - Caddy serves IP addresses and local/internal hostnames over HTTPS with locally-trusted certificates. Examples: `localhost`, `127.0.0.1`.
 - Caddy serves public DNS names over HTTPS with certificates from [Let's Encrypt](https://letsencrypt.org). Examples: `example.com`, `sub.example.com`, `*.example.com`.
 
-Caddy keeps all certificates renewed, and redirects HTTP (default port 80) to HTTPS (default port 443) automatically.
+Caddy keeps all certificates renewed, and redirects HTTP (default port 80) to HTTPS (default port 443) automatically (provided [Activation](/docs/automatic-https#activation) is successful).
 
 **For local HTTPS:**
 
@@ -75,6 +75,8 @@ When automatic HTTPS is activated, the following occurs:
 Automatic HTTPS never overrides explicit configuration.
 
 You can [customize or disable automatic HTTPS](/docs/json/apps/http/servers/automatic_https/) if necessary.
+
+<aside class="tip">Disabling redirects keeps the HTTPS port enabled as the default port implicitly assigned to an address. The HTTP port was only binded for redirects, to support connecting to both ports, you must [explicitly listen on both for an address](/docs/automatic-https#examples).</aside>
 
 
 ## Hostname requirements
@@ -220,3 +222,83 @@ Before attempting any ACME transactions, Caddy will test the configured storage 
 Caddy can obtain and manage wildcard certificates when it is configured to serve a site with a qualifying wildcard name. A site name qualifies for a wildcard if only its left-most domain label is a wildcard. For example, `*.example.com` qualifies, but these do not: `sub.*.example.com`, `foo*.example.com`, `*bar.example.com`, and `*.*.example.com`.
 
 To get a wildcard from Let's Encrypt, you simply need to enable the [DNS challenge](#dns-challenge) and use a wildcard domain in your config. We recommend using wildcards only when you have so many subdomains that you would encounter CA rate limits trying to obtain certificates for them all.
+
+
+
+## Examples
+
+Caddy implicitly uses the HTTPS port (default 443) for your [server addresses](/docs/conventions#network-addresses) that don't assign an explicit port (which would disable automatic HTTPS). 
+
+The global setting `auto_https` has two values:
+- `disable_redirects` adds an implicit HTTP port redirect.
+- `off` disables automatic HTTPS, default implicit port changes to HTTP for all server addresses.
+
+| auto_https        | HTTP        | HTTPS       |
+|-------------------|-------------|-------------|
+| *default*         | redirects   | connects    |
+| disable_redirects | unreachable | connects    |
+| off               | connects    | unreachable |
+
+`unreachable` is the default without additional configuration when using the related `auto_https` setting.
+`connects` is the implicitly assigned port.
+
+
+### Disabling automatic HTTPS
+
+For local development environments, you can prevent serving via HTTPS by providing an explicit port assignment(disable per server address) or using the global `auto_https off` setting which will change Caddy's implicit port to be the HTTP port (default 80) globally.
+
+```caddy
+{
+  auto_https off
+}
+
+# Have caddy implicitly use the HTTP port
+localhost {
+  root * /usr/share/caddy
+
+  file_server
+}
+```
+
+```caddy
+# Alternatively, provide an explicit port
+localhost:9000 {
+  root * /usr/share/caddy
+
+  file_server
+}
+```
+
+
+### HTTP and HTTPS without redirect
+
+If you want to serve content through both HTTP and HTTPS ports without HTTP redirects, your server address cannot rely on a single implicit port and you must explicitly declare the intent to listen from both ports.
+
+To do so, you can [map several addresses to a site block as a list](/docs/caddyfile/concepts#addresses) separated with `,`:
+
+```caddy
+{
+  auto_https disable_redirects
+}
+
+# Uses the HTTP and HTTPS by protocol
+# These are configurable as global settings
+http://localhost, https://localhost {
+  root * /usr/share/caddy
+
+  file_server
+}
+```
+
+```caddy
+{
+  auto_https disable_redirects
+}
+
+# Alternatively provide explicit ports
+localhost:80, localhost:443 {
+  root * /usr/share/caddy
+
+  file_server
+}
+```
--- a/src/docs/markdown/quick-starts/https.md
+++ b/src/docs/markdown/quick-starts/https.md
@@ -7,7 +7,9 @@ title: HTTPS quick-start
 This guide will show you how to get up and running with [fully-managed HTTPS](/docs/automatic-https) in no time.
 
 <aside class="tip">
-	Caddy uses HTTPS for all sites by default, as long as a host name is provided in the config. This tutorial assumes you want to get a publicly-trusted site (i.e. not "localhost") up over HTTPS, so we'll be using a public domain name and external ports.
+	Caddy uses HTTPS for all sites by default, as long as a host name is provided in the config. This tutorial assumes you want to get a publicly-trusted site (i.e. not "localhost") up over HTTPS, so we'll be using a public domain name and external ports. 
+
+	For a "localhost" setup, see examples [here](/docs/automatic-https#examples).
 </aside>
 
 **Prerequisites:**
--- a/src/docs/markdown/caddyfile/options.md
+++ b/src/docs/markdown/caddyfile/options.md
@@ -60,4 +60,4 @@ Possible options are:
 	- **interval** and **burst** allows `<n>` certificate operations within `<duration>` interval.
 - **local_certs** causes all certificates to be issued internally by default, rather than through a (public) ACME CA such as Let's Encrypt. This is useful in development environments.
 - **key_type** specifies the type of key to generate for TLS certificates; only change this if you have a specific need to customize it.
-- **auto_https** configure automatic HTTPS. It can either disable it entirely (`off`) or disable only HTTP-to-HTTPS redirects (`disable_redirects`).
+- **auto_https** configure automatic HTTPS. It can either disable it entirely (`off`) or disable only HTTP-to-HTTPS redirects (`disable_redirects`). [Examples](/docs/automatic-https#examples).


src/docs/markdown/automatic-https.md
```diff
@@ -23,7 +23,7 @@ Here's a 28-second video showing how it works:
 - Caddy serves IP addresses and local/internal hostnames over HTTPS with locally-trusted certificates. Examples: `localhost`, `127.0.0.1`.
 - Caddy serves public DNS names over HTTPS with certificates from [Let's Encrypt](https://letsencrypt.org). Examples: `example.com`, `sub.example.com`, `*.example.com`.
 
-Caddy keeps all certificates renewed, and redirects HTTP (default port 80) to HTTPS (default port 443) automatically.
+Caddy keeps all certificates renewed, and redirects HTTP (default port 80) to HTTPS (default port 443) automatically (provided [Activation](/docs/automatic-https#activation) is successful).
```
francislavoie:
```suggestion
Caddy keeps all certificates renewed, and redirects HTTP (default port 80) to HTTPS (default port 443) automatically, provided that [activation](/docs/automatic-https#activation) is successful.
```

src/docs/markdown/automatic-https.md
```diff
@@ -76,6 +76,8 @@ Automatic HTTPS never overrides explicit configuration.
 
 You can [customize or disable automatic HTTPS](/docs/json/apps/http/servers/automatic_https/) if necessary.
 
+<aside class="tip">Disabling redirects keeps the HTTPS port enabled as the default port implicitly assigned to an address. The HTTP port was only binded for redirects, to support connecting to both ports, you must [explicitly listen on both for an address](/docs/automatic-https#examples).</aside>
```
francislavoie:
```suggestion
<aside class="tip">Disabling redirects keeps the HTTPS port enabled as the default port implicitly assigned to an address. The HTTP port is only bound for redirects and for the ACME HTTP challenge. If you need to serve over HTTP, you must explicitly configure Caddy to do it. [See the examples](/docs/automatic-https#examples).</aside>
```

src/docs/markdown/automatic-https.md
```diff
@@ -220,3 +222,83 @@ Before attempting any ACME transactions, Caddy will test the configured storage
...
+`connects` is the implicitly assigned port.
+
+
+### Disabling automatic HTTPS
+
+For local development environments, you can prevent serving via HTTPS by providing an explicit port assignment(disable per server address) or using the global `auto_https off` setting which will change Caddy's implicit port to be the HTTP port (default 80) globally.
+
+```caddy
+{
+  auto_https off
```
francislavoie: In general we use tabs for Caddyfile config

```suggestion
	auto_https off
```

src/docs/markdown/automatic-https.md
```diff
@@ -220,3 +222,83 @@ Before attempting any ACME transactions, Caddy will test the configured storage
...
+}
+
+# Have caddy implicitly use the HTTP port
+localhost {
+  root * /usr/share/caddy
+
+  file_server
```
francislavoie:
```suggestion
	root * /usr/share/caddy

	file_server
```

src/docs/markdown/automatic-https.md
```diff
@@ -220,3 +222,83 @@ Before attempting any ACME transactions, Caddy will test the configured storage
...
+}
+```
+
+```caddy
+# Alternatively, provide an explicit port
+localhost:9000 {
+  root * /usr/share/caddy
+
+  file_server
```
francislavoie:
```suggestion
	root * /usr/share/caddy

	file_server
```

src/docs/markdown/automatic-https.md
```diff
@@ -220,3 +222,83 @@ Before attempting any ACME transactions, Caddy will test the configured storage
 Caddy can obtain and manage wildcard certificates when it is configured to serve a site with a qualifying wildcard name. A site name qualifies for a wildcard if only its left-most domain label is a wildcard. For example, `*.example.com` qualifies, but these do not: `sub.*.example.com`, `foo*.example.com`, `*bar.example.com`, and `*.*.example.com`.
 
 To get a wildcard from Let's Encrypt, you simply need to enable the [DNS challenge](#dns-challenge) and use a wildcard domain in your config. We recommend using wildcards only when you have so many subdomains that you would encounter CA rate limits trying to obtain certificates for them all.
+
+
+
+## Examples
+
+Caddy implicitly uses the HTTPS port (default 443) for your [server addresses](/docs/conventions#network-addresses) that don't assign an explicit port (which would disable automatic HTTPS).
```
francislavoie:
```suggestion
Caddy implicitly uses the HTTPS port (default 443) for your [server addresses](/docs/conventions#network-addresses) that don't specify a port explicitly (which would disable automatic HTTPS). 
```

polarathene:
Would it be appropriate to also briefly touch on [this snippet](https://github.com/caddyserver/caddy/issues/3636#issuecomment-668967388):

```
http://localhost {
	redir https://localhost{uri}
}
```

Which was explained to me as roughly what Automatic HTTPS achieves? (I thought that it was doing more HTTPS stuff, but I guess that's separate from the automatic feature).

src/docs/markdown/automatic-https.md
```diff
@@ -220,3 +222,83 @@ Before attempting any ACME transactions, Caddy will test the configured storage
...
 
 To get a wildcard from Let's Encrypt, you simply need to enable the [DNS challenge](#dns-challenge) and use a wildcard domain in your config. We recommend using wildcards only when you have so many subdomains that you would encounter CA rate limits trying to obtain certificates for them all.
+
+
+
+## Examples
+
+Caddy implicitly uses the HTTPS port (default 443) for your [server addresses](/docs/conventions#network-addresses) that don't assign an explicit port (which would disable automatic HTTPS). 
+
+The global setting `auto_https` has two values:
```
francislavoie:
```suggestion
Automatic HTTPS can be configured via the Caddyfile with the [`auto_https` global option](/docs/caddyfile/options), or via [per-server JSON configuration](/docs/json/apps/http/servers/automatic_https/). The Caddyfile option can be set to either of the following:
```

src/docs/markdown/automatic-https.md
```diff
@@ -220,3 +222,83 @@ Before attempting any ACME transactions, Caddy will test the configured storage
...
+- `disable_redirects` adds an implicit HTTP port redirect.
```
francislavoie:
```suggestion
- `disable_redirects` which disables the implicit HTTP->HTTPS redirect.
```

src/docs/markdown/automatic-https.md
```diff
@@ -220,3 +222,83 @@ Before attempting any ACME transactions, Caddy will test the configured storage
...
+- `off` disables automatic HTTPS, default implicit port changes to HTTP for all server addresses.
```
francislavoie:
```suggestion
- `off` which disables automatic HTTPS altogether, including the HTTP->HTTPS redirect and automatic enabling of TLS for sites that meet the requirements.
```

src/docs/markdown/automatic-https.md
```diff
@@ -220,3 +222,83 @@ Before attempting any ACME transactions, Caddy will test the configured storage
...
+| disable_redirects | unreachable | connects    |
+| off               | connects    | unreachable |
+
+`unreachable` is the default without additional configuration when using the related `auto_https` setting.
+`connects` is the implicitly assigned port.
+
+
+### Disabling automatic HTTPS
+
+For local development environments, you can prevent serving via HTTPS by providing an explicit port assignment(disable per server address) or using the global `auto_https off` setting which will change Caddy's implicit port to be the HTTP port (default 80) globally.
```
francislavoie:
```suggestion
For local development environments, you can prevent serving via HTTPS by either specifying `http://` or providing a non-HTTPS port to disable per site, or by adding `auto_https off` to the Caddyfile global options which will change Caddy's implicit port to be the HTTP port (default 80).
```

src/docs/markdown/automatic-https.md
```diff
@@ -220,3 +222,83 @@ Before attempting any ACME transactions, Caddy will test the configured storage
...
+
+```caddy
+{
+  auto_https off
+}
+
+# Have caddy implicitly use the HTTP port
```
francislavoie: I think the comments should be hoisted out of the code blocks with a quick line right above them that describe what follows. Reads better to have the purpose of the example just before it. Also reads better to have a line of text between each example code block.

src/docs/markdown/automatic-https.md
```diff
@@ -220,3 +222,83 @@ Before attempting any ACME transactions, Caddy will test the configured storage
...
+
+### HTTP and HTTPS without redirect
+
+If you want to serve content through both HTTP and HTTPS ports without HTTP redirects, your server address cannot rely on a single implicit port and you must explicitly declare the intent to listen from both ports.
+
+To do so, you can [map several addresses to a site block as a list](/docs/caddyfile/concepts#addresses) separated with `,`:
+
+```caddy
+{
+  auto_https disable_redirects
```
francislavoie:
```suggestion
	auto_https disable_redirects
```

src/docs/markdown/automatic-https.md
```diff
@@ -220,3 +222,83 @@ Before attempting any ACME transactions, Caddy will test the configured storage
...
+}
+
+# Uses the HTTP and HTTPS by protocol
+# These are configurable as global settings
+http://localhost, https://localhost {
+  root * /usr/share/caddy
+
+  file_server
```
francislavoie:
```suggestion
	root * /usr/share/caddy

	file_server
```

src/docs/markdown/automatic-https.md
```diff
@@ -220,3 +222,83 @@ Before attempting any ACME transactions, Caddy will test the configured storage
...
+}
+```
+
+```caddy
+{
+  auto_https disable_redirects
```
francislavoie:
```suggestion
	auto_https disable_redirects
```

src/docs/markdown/automatic-https.md
```diff
@@ -220,3 +222,83 @@ Before attempting any ACME transactions, Caddy will test the configured storage
...
+}
+
+# Alternatively provide explicit ports
+localhost:80, localhost:443 {
+  root * /usr/share/caddy
+
+  file_server
```
francislavoie:
```suggestion
	root * /usr/share/caddy

	file_server
```

src/docs/markdown/automatic-https.md
```diff
@@ -220,3 +222,83 @@ Before attempting any ACME transactions, Caddy will test the configured storage
...
+
+To do so, you can [map several addresses to a site block as a list](/docs/caddyfile/concepts#addresses) separated with `,`:
+
+```caddy
+{
+  auto_https disable_redirects
+}
+
+# Uses the HTTP and HTTPS by protocol
+# These are configurable as global settings
```
francislavoie: Same comment applies for these examples, move these comments out of the code blocks.

src/docs/markdown/automatic-https.md
```diff
@@ -220,3 +222,83 @@ Before attempting any ACME transactions, Caddy will test the configured storage
...
+  file_server
+}
+```
+
+
+### HTTP and HTTPS without redirect
+
+If you want to serve content through both HTTP and HTTPS ports without HTTP redirects, your server address cannot rely on a single implicit port and you must explicitly declare the intent to listen from both ports.
+
+To do so, you can [map several addresses to a site block as a list](/docs/caddyfile/concepts#addresses) separated with `,`:
```
francislavoie:
```suggestion
If you wish to serve content through both HTTP and HTTPS, without HTTP->HTTPS redirects, your site label cannot must explicitly declare the intent to listen on both ports. To do so, [specify multiple site labels](/docs/caddyfile/concepts#addresses), separated by a comma or whitespace.
```

francislavoie (changes requested):

src/docs/markdown/automatic-https.md
polarathene: > site label 

I did not see this terminology used in the docs yet? I've seen "Network Addresses", "Addresses", and their corresponding Block "site block". I do prefer site label though, in a few places I've prepended "server" when mentioning addresses for added clarity. Since it's not always for websites(could be for an API), is "service address" better?

[Conventions - Network Addresses](https://caddyserver.com/docs/conventions#network-addresses) tip section defines "Network Addresses" pretty well, additionally the [concepts section](https://caddyserver.com/docs/caddyfile/concepts#addresses) visual breakdown has a legend calling it a "site address", and in the "Addresses" section wildcard matching refers to a label as part of a hostname.

[URI](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier) also seems to be an apt description I think?

> cannot must

I think the "cannot" was meant to be removed?

src/docs/markdown/automatic-https.md
polarathene: Not worth mentioning here the implicit default port being switched from HTTPS to HTTP? I guess that's still covered in the table directly below.

src/docs/markdown/automatic-https.md
polarathene: I have pushed changes removing the "cannot", and renamed "site label" and "server addresses" to "site address/es".

src/docs/markdown/automatic-https.md
francislavoie: Technically it doesn't do that.

```
{
        auto_https off
}

localhost

respond "Hello World"
```

Adapts to:

```json
{
  "apps": {
    "http": {
      "servers": {
        "srv0": {
          "listen": [
            ":443"
          ],

          "routes": [
            {
              "match": [
                {
                  "host": [
                    "localhost"
                  ]
                }
              ],
              "handle": [
                {
                  "handler": "subroute",
                  "routes": [
                    {
                      "handle": [
                        {
                          "body": "Hello World",
                          "handler": "static_response"
                        }
                      ]
                    }
                  ]
                }
              ],
              "terminal": true
            }
          ],
          "tls_connection_policies": [
            {}
          ],
          "automatic_https": {
            "disable": true
          }
        }
      }
    }
  }
}
```

It still only listens to `:443` but TLS will not be automatically managed (so connections will fail). So turning off `auto_https` in that situation only really makes sense if you also manually configure TLS (with your own cert/key for example).

src/docs/markdown/automatic-https.md
francislavoie: There's an unpublished `spec.md` document that you can find in here that talks a lot about labels. That was the terminology used in Caddy v1 and I may have missed the decision where @mholt switched the terminology... but I'm not sure.

As you said, the wildcard bit specifically talks about one segment of the address being a label, which follows from the `{labels.*}` placeholders that exist. Maybe label isn't the best word here.

👍 for the "cannot" typo

src/docs/markdown/automatic-https.md
```diff
@@ -23,7 +23,7 @@ Here's a 28-second video showing how it works:
 - Caddy serves IP addresses and local/internal hostnames over HTTPS with locally-trusted certificates. Examples: `localhost`, `127.0.0.1`.
 - Caddy serves public DNS names over HTTPS with certificates from [Let's Encrypt](https://letsencrypt.org). Examples: `example.com`, `sub.example.com`, `*.example.com`.
 
-Caddy keeps all certificates renewed, and redirects HTTP (default port 80) to HTTPS (default port 443) automatically.
+Caddy keeps all certificates renewed, and redirects HTTP (default port 80) to HTTPS (default port 443) automatically, provided that [activation](/docs/automatic-https#activation) is successful.
```
mholt, line 26: This is basically the same thing as "Automatic HTTPS works as long as it does not fail," which seems obvious/redundant: of course if something fails, it will not succeed. I think we actually want to convey, "Auto HTTPS has certain criteria for activation."

How about a different way of introducing the activation criteria...

```suggestion
Provided that [auto-HTTPS activates](/docs/automatic-https#activation), Caddy keeps all certificates renewed, and redirects HTTP (default port 80) to HTTPS (default port 443) automatically.
```

src/docs/markdown/automatic-https.md
```diff
@@ -76,6 +76,8 @@ Automatic HTTPS never overrides explicit configuration.
 
 You can [customize or disable automatic HTTPS](/docs/json/apps/http/servers/automatic_https/) if necessary.
 
+<aside class="tip">Disabling redirects keeps the HTTPS port enabled as the default port implicitly assigned to an address. The HTTP port is only bound for redirects and for the ACME HTTP challenge. If you need to serve over HTTP, you must explicitly configure Caddy to do it. [See the examples](/docs/automatic-https#examples).</aside>
```
mholt, line 79: This is confusing because this section is not about disabling auto-HTTPS (that is below). Even I had to read this several times before it started to make sense, and even now I'm not sure what "need to serve over HTTP" means. Serving redirects? Serving content? Anyway, I think this should be moved or removed.

src/docs/markdown/automatic-https.md
```diff
@@ -220,3 +222,84 @@ Before attempting any ACME transactions, Caddy will test the configured storage
 Caddy can obtain and manage wildcard certificates when it is configured to serve a site with a qualifying wildcard name. A site name qualifies for a wildcard if only its left-most domain label is a wildcard. For example, `*.example.com` qualifies, but these do not: `sub.*.example.com`, `foo*.example.com`, `*bar.example.com`, and `*.*.example.com`.
 
 To get a wildcard from Let's Encrypt, you simply need to enable the [DNS challenge](#dns-challenge) and use a wildcard domain in your config. We recommend using wildcards only when you have so many subdomains that you would encounter CA rate limits trying to obtain certificates for them all.
+
+
+
+## Examples
+
+Caddy implicitly uses the HTTPS port (default 443) for your [site addresses](/docs/conventions#network-addresses) that don't specify a port explicitly (which would disable automatic HTTPS).
```
mholt, line 230: This isn't accurate. Any port other than 80 will still use HTTPS by default.

src/docs/markdown/automatic-https.md
```diff
@@ -220,3 +222,84 @@ Before attempting any ACME transactions, Caddy will test the configured storage
...
 To get a wildcard from Let's Encrypt, you simply need to enable the [DNS challenge](#dns-challenge) and use a wildcard domain in your config. We recommend using wildcards only when you have so many subdomains that you would encounter CA rate limits trying to obtain certificates for them all.
+
+
+
+## Examples
+
+Caddy implicitly uses the HTTPS port (default 443) for your [site addresses](/docs/conventions#network-addresses) that don't specify a port explicitly (which would disable automatic HTTPS). 
+
+Automatic HTTPS can be configured via the Caddyfile with the [`auto_https` global option](/docs/caddyfile/options), or via [per-server JSON configuration](/docs/json/apps/http/servers/automatic_https/). The Caddyfile option can be set to either of the following:
+- `disable_redirects` which disables the implicit HTTP->HTTPS redirect.
```
mholt, line 234:
```suggestion

- `disable_redirects` which disables the implicit HTTP->HTTPS redirect.
```

src/docs/markdown/automatic-https.md
```diff
@@ -220,3 +222,84 @@ Before attempting any ACME transactions, Caddy will test the configured storage
...
+- `off` which disables automatic HTTPS altogether, including the HTTP->HTTPS redirect and automatic enabling of TLS for sites that meet the requirements.
+
+| auto_https        | HTTP        | HTTPS       |
+|-------------------|-------------|-------------|
+| *default*         | redirects   | connects    |
+| disable_redirects | unreachable | connects    |
+| off               | connects    | unreachable |
```
mholt, line 241: What conditions and configuration does this table assume? For example, here's a (contrived) Caddyfile where the table is not accurate:

```
{
    auto_https off
}

https://example.com

tls cert.pem key.pem
```

This will connect on port 443 but not 80. (Granted, this Caddyfile uses the `auto_https` option unnecessarily since providing my own cert and key disables it for that site anyway.)

Also a slight formatting suggestion, since `auto_https` refers to a config parameter:

```suggestion
| `auto_https`      | HTTP        | HTTPS       |
|-------------------|-------------|-------------|
| *default*         | redirects   | connects    |
| disable_redirects | unreachable | connects    |
| off               | connects    | unreachable |
```

src/docs/markdown/automatic-https.md
```diff
@@ -220,3 +222,84 @@ Before attempting any ACME transactions, Caddy will test the configured storage
...
+
+`unreachable` is the default without additional configuration when using the related `auto_https` setting.
```
mholt, line 243: What does this mean though? Shouldn't this describe how the socket is not opened implicitly?

Also I'm not sure "unreachable" is the right word. Whether a socket is reachable is external and unrelated to the web server: I think you want "does not connect" or something like that. Port 443 (or whatever) might very well be reachable, but not able to connect.

src/docs/markdown/automatic-https.md
```diff
@@ -220,3 +222,84 @@ Before attempting any ACME transactions, Caddy will test the configured storage
...
+`connects` is the implicitly assigned port.
```
mholt, line 244: This is also confusing. "Connects" is not a port. This should describe what connects means.

src/docs/markdown/automatic-https.md
```diff
@@ -220,3 +222,84 @@ Before attempting any ACME transactions, Caddy will test the configured storage
...
+
+
+### Disabling automatic HTTPS
+
+For local development environments, you can prevent serving via HTTPS by either specifying `http://` or providing a non-HTTPS port to disable per site, or by adding `auto_https off` to the Caddyfile global options which will change Caddy's implicit port to be the HTTP port (default 80).
```
mholt, line 249: This works in _any_ environment, not just local development.

mholt: I would like to see this section refer to the activation criteria and then show 1:1 how _not_ meeting those criteria disables auto HTTPS.

src/docs/markdown/automatic-https.md
```diff
@@ -220,3 +222,84 @@ Before attempting any ACME transactions, Caddy will test the configured storage
 Caddy can obtain and manage wildcard certificates when it is configured to serve a site with a qualifying wildcard name. A site name qualifies for a wildcard if only its left-most domain label is a wildcard. For example, `*.example.com` qualifies, but these do not: `sub.*.example.com`, `foo*.example.com`, `*bar.example.com`, and `*.*.example.com`.
 
 To get a wildcard from Let's Encrypt, you simply need to enable the [DNS challenge](#dns-challenge) and use a wildcard domain in your config. We recommend using wildcards only when you have so many subdomains that you would encounter CA rate limits trying to obtain certificates for them all.
+
+
+
+## Examples
```
mholt, line 228: I think this section is a good idea, but it's only Caddyfile examples. The Caddyfile is just a config adapter. This section should ideally be moved to an area of the docs specifically about the Caddyfile.

src/docs/markdown/automatic-https.md
```diff
@@ -220,3 +222,84 @@ Before attempting any ACME transactions, Caddy will test the configured storage
...
+
+Alternatively, provide an explicit port:
+
+```caddy
+localhost:9000 {
+	root * /usr/share/caddy
+
+	file_server
+}
+```
```
mholt, line 273: As mentioned above, this is wrong; specifying a port does not disable automatic HTTPS. Read the criteria closely, and make sure to try every config you add to the docs to prove it works.

src/docs/markdown/automatic-https.md
```diff
@@ -220,3 +222,84 @@ Before attempting any ACME transactions, Caddy will test the configured storage
...
+
+
+### HTTP and HTTPS without redirect
+
+If you wish to serve content through both HTTP and HTTPS, without HTTP->HTTPS redirects, your site address must explicitly declare the intent to listen on both ports. To do so, [specify multiple site labels](/docs/caddyfile/concepts#addresses), separated by a comma or whitespace.
+
+HTTP and HTTPS by protocol, default ports can be configured via the Caddyfile [`http_port` and `https_port` global options](/docs/caddyfile/options):
```
mholt: This sentence isn't one... try this instead:

```suggestion
One way to do this is by specifying the protocol scheme, which infers the standard ports (the respective ports can be customized using [`http_port` and `https_port` global options](/docs/caddyfile/options)):
```

src/docs/markdown/quick-starts/https.md
```diff
@@ -7,7 +7,9 @@ title: HTTPS quick-start
 This guide will show you how to get up and running with [fully-managed HTTPS](/docs/automatic-https) in no time.
 
 <aside class="tip">
-	Caddy uses HTTPS for all sites by default, as long as a host name is provided in the config. This tutorial assumes you want to get a publicly-trusted site (i.e. not "localhost") up over HTTPS, so we'll be using a public domain name and external ports.
+	Caddy uses HTTPS for all sites by default, as long as a host name is provided in the config. This tutorial assumes you want to get a publicly-trusted site (i.e. not "localhost") up over HTTPS, so we'll be using a public domain name and external ports. 
+
+	For a "localhost" setup, see examples [here](/docs/automatic-https#examples).
```
mholt, line 12: This is HTML, so the extra line break here won't render (at least, I think).

This is confusing though, since I don't know what a `"localhost" setup` means. A site using `localhost` as the site name? A local dev environment? I think it is ambiguous.

mholt (changes requested):
Thanks for contributing to the docs! This'll need some work before it's acceptable, but I like some of where it's going.

> Not sure where these JSON docs are located, I take it they're not markdown?

They're dynamically generated from a documentation server backend; see https://github.com/caddyserver/website/blob/master/src/docs/json/index.html and https://github.com/caddyserver/moduledoc.

src/docs/markdown/automatic-https.md
polarathene: This is why the docs need improving :+1: 

Although that does make the rule a little confusing. If only an explicit port 80 opts out of HTTPS for a site block, is that really the right way to go about it? (probably too late to change now since v2 is out)

If I add a different port for a service and want to opt-out of HTTPS being implicitly added, how do I do that? Eg SMTP implicit(465) and explicit(587) TLS ports where HTTPS isn't desired?

src/docs/markdown/automatic-https.md
polarathene: The `unreachable` and `connects` are defined beneath the table. They clarify that default without additional explicit configuration (such as a protocol scheme or port on a site address) `connects` is the ~implicit port/protocol used, and `unreachable` is just that, not implicitly reachable~(**EDIT:** Misunderstood this, was sure I had tested or seen in the docs of the default implicit port changing between HTTP/HTTPS, but I'm mistaken). `redirects` wasn't explained as I think that's been covered clearly that it is just redirecting HTTP connections to HTTPS to be handled.

> For example, here's a (contrived) Caddyfile where the table is not accurate

You didn't rely on the site address being implicitly configured ~for HTTP~(**EDIT: Caddy doesn't do that)/HTTPS, so it's not valid. Granted I've only thought of it regarding ports.

There's an example for `disable_redirects` that uses the protocol schemes, but I personally had no idea how HTTPS is configured without the auto feature or what `auto_https` does "under the hood" equivalent to manual configuration. Traefik has a nice way with entrypoints and routers, the docs for Caddy probably touch on this somewhere, I just haven't gone through them in a linear fashion. 

Perhaps it's worth mentioning early on in the doc (via a link to where ever non-auto equivalents are explained) and maybe once more at the start of the examples section?

Does using an explicit port 443 also enable HTTPS(provided it matches the configured HTTPS port)? What does a user need to do to enable TLS / HTTPS on a non-standard port that's also not the standard HTTPS configured port(if changed from the default 443)? (**EDIT:** HTTPS/443 is always the default implicit port, even with `auto_https off` just that drops TLS which raises an error response, using any port other than 80 is treated as HTTPS, so the question changes to opting out of TLS for non-standard ports.)

> Also a slight formatting suggestion, since auto_https refers to a config parameter

The values in that column are also relevant to use inline code blocks then? (except for *default*, not sure if that's presented as a non-value, does it have any valid value like `on`?)

src/docs/markdown/automatic-https.md
polarathene: There is mention of disabling it in that section, it's a section I viewed early on IIRC as it was a rather lengthy document and not all the information seemed important to understand what to do (or what I was doing wrong). It's been isolated as a tip to separate the context a little for the reason you bring up, that was meant to be less confusing actually..

It's about `disable_redirects` (which both Caddyfile and JSON support for config), so not about serving redirects, I could change to "serve content" unless there is a better way to phrase it(expose HTTP port, be reachable over the HTTP port, support connections over HTTP, etc).

I didn't have the bit about ACME HTTP challenge there, that was added as a suggestion to clarify that the HTTP port was implicitly redirecting to HTTPS in addition to being used for that challenge.

Examples section was linked to for more information on how to configure Caddy properly when using `disable_redirects`, so that you could access both HTTP and HTTPS, as this was not very straightforward.. I was originally under the belief that HTTP was the default implicit port, and that the `auto_https` feature was implicitly adding HTTPS support/connections, such that `disable_redirects` would implicitly still handle both ports, but do as the value states and just disable HTTP->HTTPS redirects.. while `off` would do that and not implicitly add HTTPS(**EDIT:** Mistaken, HTTP isn't the implicit port for `off` and HTTPS is still used implicitly, just without TLS thus it fails).

src/docs/markdown/automatic-https.md
polarathene: > This is basically the same thing as "Automatic HTTPS works as long as it does not fail," which seems obvious/redundant

Mostly the intent was to point the user to the Activation section for a list of criteria for when `auto_https` would work or fail. The original statement just implies it all works well with no clarification of why it might fail.

A site address of `localhost` by default implicitly connects over HTTPS with HTTP->HTTPS redirect, `off` connects to ~HTTP with no HTTPS~(**EDIT:** I'm mistaken, still implicitly `443` but refuses to serve) or redirect, while `disable_redirects` results in HTTP not being reachable(not my original expectation), or `localhost:80` being reachable but no HTTPS anymore despite `auto_https disable_redirects` behaviour without the explicit port, and that turns out to be due to another non-obvious rule/condition at play.

So while it seems redundant, it's aimed at directing the user to the right docs when troubleshooting, they shouldn't need to go through the entire docs page looking for some hopeful explanation or worse trying to find such information elsewhere in the docs, especially if this feature spans multiple doc pages about it (which might not be known of at the time either).

---

Slight change of phrasing, as "Activation" personally didn't *click* as well when I first saw it. "rules" might also be appropriate alternative to "conditions":

```suggestion
When the [conditions for automatic HTTPS](/docs/automatic-https#activation) are valid, Caddy keeps all certificates renewed, and redirects HTTP (default port 80) to HTTPS (default port 443) automatically.
```

src/docs/markdown/automatic-https.md
polarathene: > This works in any environment, not just local development.

Yes, I could probably have re-worded that better. I had seen the quick-starts guide earlier that was more involved in config/requirements IIRC, and explicitly stated that it wasn't for local environment(eg `localhost` site address). I may have made this mistake here due to train of thought with the localhost examples used, and originally having `local_certs` global option set which I apparently didn't need to do for localhost site address and later removed before committing.

> I would like to see this section refer to the activation criteria and then show 1:1 how not meeting those criteria disables auto HTTPS.

Sure I can do that :+1:

src/docs/markdown/automatic-https.md
polarathene: As long as it's referenced from this doc for discovery, sure.

Most of the docs snippets I've seen are for Caddyfile configs, and as a new user I've found that most useful. I haven't tried the JSON configs yet, I'm not sure how much that differs but I have noticed it doesn't seem to be 1:1 equivalent, such as the JSON docs for settings of auto HTTPS feature has fields like `skip`.

Personally I'd like the tabbed example snippets like Traefik has, which lets you view the config in various supported formats. I was interested in the TOML adapter, but all I found was a link to a github repo with no documentation on how to use it, so I just stuck with Caddyfile for now.

If I move these, do you have any suggestions for how you want that handled? Is it going to become a large examples page for Caddyfile with many sections for different features over time, as a sort of reference? Or sub-pages for each feature?

src/docs/markdown/quick-starts/https.md
polarathene: I was trying to refer to the earlier mention of [`"localhost"` in that very doc snippet](https://caddyserver.com/docs/quick-starts/https):

> This tutorial assumes you want to get a publicly-trusted site (i.e. not "localhost") up over HTTPS

Perhaps "setup" is not the right word and doesn't work with that prior context I've quoted. I assumed it meant for a "site address"(URI?) that uses `localhost` and would be accessed as such locally. How should I better clarify that?

src/docs/markdown/automatic-https.md
polarathene: ~Should the table be revised to indicate instead which protocol scheme/port is implicitly used?~ In which case should `redirects` be changed/clarified or left as is?

**EDIT:** Misunderstood how the implicit handling works. There is no implicit HTTP other than the default redirect to HTTPS. This table is not useful then..

src/docs/markdown/automatic-https.md
polarathene: This was my fault for only thinking of site addresses regarding ports(rather than Network Addresses), would it be more accurate to state that Caddy is ~implicitly adding the protocol scheme~(**EDIT:** it's always implicitly https/443 afaik? even with `auto_https off`), which infers the port from `http_port` and `https_port` or their defaults?

The protocol scheme threw me off a bit, as while the http/https ports can be configured for Caddy to use different ports, the protocol schemes felt more explicit about what ports they'd refer to, but I guess they too match `http_port`/`https_port` and aren't related to the user connecting over `http/https` via their browser per se?

~I can drop these with changing the table to focus more on which protocol/port is used implicitly instead.~

src/docs/markdown/automatic-https.md
polarathene: @francislavoie ah ok... thanks for clarifying that. I could have sworn I had tested it previously and had HTTP work implicitly, it seems I'm mistaken. Tried looking for mentions about this in the docs, but again can't seem to find that information I thought I had read of the default implicit port changing.

So the default port is always implicitly `443`? (Unless some other explicit configuration like a protocol scheme is provided)

I should probably clarify that in the docs if its not already.

src/docs/markdown/automatic-https.md
polarathene: Just tried a config without touching `auto_https`, used `localhost:8000` and connected to it via `curl`, but it responded with:

> Client sent an HTTP request to an HTTPS server.

So HTTPS (or just TLS?) by default. How would I opt out of that for a site block address while other site blocks leverage auto HTTPS? 

This is important if Caddy is handling the ports for IMAP service that for whatever reason makes plaintext port available to connect to (which can leverage StartTLS). A better example might be the SMTP port 25, which is required to be reachable for incoming mail traffic and is plaintext but can leverage StartTLS to negotiate an explicit TLS connection, something Caddy probably doesn't support and would need to just proxy/passthrough the connection to the SMTP service.

src/docs/markdown/automatic-https.md
mholt: "Conditions are valid" is weird/awkward. I still prefer my suggestion above, as "activation" is probably the best word for it.

src/docs/markdown/automatic-https.md
mholt: Let's remove this paragraph, it's just too confusing and out of place here.

src/docs/markdown/automatic-https.md
mholt: "Conditions are valid" is weird/awkward. I still prefer my suggestion above, as "activation" is probably the best word for it.

src/docs/markdown/automatic-https.md
mholt: I think that examples and other config tutorials that aren't central to the core of Caddy are a perfect fit for our wiki: https://caddy.community/c/wiki/13 -- that'd probably be the best place for this content for now. We can always bring stuff into the official docs later, but taking things out is difficult (and must reach a high standard of quality and structure, organization, and placement).

src/docs/markdown/automatic-https.md
mholt: > If only an explicit port 80 opts out of HTTPS for a site block,

That's not quite what I said; any of the other conditions listed on this page will _also_ deactivate auto HTTPS (for example, no hostname in the config, i.e. a site address of `:1234`).

It's fairly nuanced and we need to be careful and rigorous about the language we use, especially in technical documentation.

> Eg SMTP implicit(465) and explicit(587) TLS ports where HTTPS isn't desired?

Caddy's HTTP server is an HTTP server, it doesn't serve SMTP; so ports for other services are irrelevant. Remember that the HTTP server is only one app module for Caddy, it can definitely have a different module for an SMTP server, with its own semantics.

> So HTTPS (or just TLS?) by default. How would I opt out of that for a site block address while other site blocks leverage auto HTTPS?

Simply break any of the conditions that activate auto-HTTPS? The existing Activation section already answers your question.

> This is important if Caddy is handling the ports for IMAP service

How (and why) are you doing that with an HTTP server?

All the talk about non-HTTP for the "automatic HTTPS" docs really makes no sense, it's totally irrelevant here. Why are we talking about mail protocols on documentation for an HTTP server?

src/docs/markdown/automatic-https.md
mholt: > The unreachable and connects are defined beneath the table. 

What I'm saying is that they don't define what they _mean_. And they don't make sense.

> Perhaps it's worth mentioning early on in the doc (via a link to where ever non-auto equivalents are explained) and maybe once more at the start of the examples section?

Mentioning what, exactly?

> Does using an explicit port 443 also enable HTTPS(provided it matches the configured HTTPS port)?

Like anything else, if it doesn't break any of the requirements here, then yes: https://caddyserver.com/docs/automatic-https#activation (so, yes)

> EDIT: HTTPS/443 is always the default implicit port

Only if auto-HTTPS activates! Which is most of the time. That's the point of Caddy's HTTP server: it's HTTPS by default.

> The values in that column are also relevant to use inline code blocks then?

This was just a personal preference, I don't care strongly either way.

src/docs/markdown/automatic-https.md
mholt: > EDIT: Misunderstood how the implicit handling works. There is no implicit HTTP other than the default redirect to HTTPS. This table is not useful then..

I think I agree on this point. The only nuance to this is that Caddy _will_ try to bind to the HTTP port to solve the ACME HTTP challenge, but this doesn't _necessarily_ happen at the same time as starting the server along with the HTTPS ports etc. Caddy still needs the HTTP port if it is to solve the HTTP challenge successfully, so it is used for more than just redirects (unless the HTTP challenge is disabled.)

But that is explained in the docs already.

We can remove the table then.

src/docs/markdown/automatic-https.md
mholt: Correct, if you specify a scheme, it uses the port configured for that scheme. By default this is, of course, 80 and 443. But in some circumstances you need to forward ports from the outside (I have to do this all the time), so you can change the ports for internal use.

src/docs/markdown/quick-starts/https.md
mholt: I think that's my question to you -- what is your examples section really about? It seems that doesn't have a clear answer at this point. I don't see any correlation to "localhost" sites and it's more about configuring auto HTTPS in the Caddyfile. Again, I think that would make a good wiki entry, but feels really out of place here.

mholt (changes requested):
Still needs some revisions if it is to be accepted. I think most of the new content would be better placed in the wiki.

src/docs/markdown/automatic-https.md
polarathene: I feel the same way about "activation" personally :)

I'll go with your suggestion, I understand the doc better now that it doesn't bother me and isn't something worth debating further.

src/docs/markdown/automatic-https.md
polarathene: So remove this whole section and add it to the wiki, or should there also be a link to these wiki examples left somewhere in the doc?

Do I need to iterate the examples on the wiki, like the PR review here, or leave it for others to contribute improvements?

mholt: > So remove this whole section and add it to the wiki

I think so. The wiki  is much easier for the community to help refine as well.

> should there also be a link to these wiki examples left somewhere in the doc?

You may absolutely link to wiki posts from the documentation, sure. :)

> Do I need to iterate the examples on the wiki, like the PR review here, or leave it for others to contribute improvements?

Why not both? 🤷‍♂️

src/docs/markdown/automatic-https.md
polarathene: > It's fairly nuanced and we need to be careful and rigorous about the language we use, especially in technical documentation.

Agreed. I think that's actually where I got confused about the implicit port changing, because when I first raised an issue about it my global options had `local_certs` set, which I later realized was not required as they're created automatically with `localhost` as the site address.

My comment about explicit port 80 was regarding ports causing an opt-out, but as you've pointed out, that's not correct either.. personally not fond of confusing that can get to initially grok, but I guess for most users it's not a problem they'd need to think about often.

> Caddy's HTTP server is an HTTP server, it doesn't serve SMTP; so ports for other services are irrelevant.
> How (and why) are you doing that with an HTTP server?
> Why are we talking about mail protocols on documentation for an HTTP server?

I'd have services like Postfix and Dovecot handling SMTP/IMAP, they expose the ports. 

Sorry, this was again me not thinking straight. As you rightly point out, we're configuring an HTTP server/service, I have seen a `reverse_proxy` directive but not given it a try yet. I was yet again thinking of alternatives to Caddy that I've used prior where I define a port upfront(which points to the service), and HTTPS especially with the port configurability had me thinking of it as generic automatic TLS with LetsEncrypt support.

I've seen a few users trying to use Caddy with their SMTP services, where Caddy acts as a reverse proxy for those ports, but manages the TLS cert renewal for a hostname (eg mail.example.com). Similar is done with Traefik, while NGINX can actually support handling StartTLS IIRC.

> All the talk about non-HTTP for the "automatic HTTPS" docs really makes no sense, it's totally irrelevant here.

Sorry, as mentioned, I was thinking of it as automatic TLS, and that the only thing that made it HTTPS was the default port and in the context of running an HTTP service.

> Simply break any of the conditions that activate auto-HTTPS? The existing Activation section already answers your question.

It wouldn't break those conditions for `mail.example.com:143`, but as you clarified, that's because this is only related to an HTTP/S service.

> Remember that the HTTP server is only one app module for Caddy, it can definitely have a different module for an SMTP server, with its own semantics.

Understood, I'll need to get more familiar with the [modules](https://caddyserver.com/docs/modules/), I guess anything not in an http namespace might not be supported by Caddyfile and require the JSON approach to configuration(or equivalent via config adapters).

src/docs/markdown/quick-starts/https.md
polarathene: I've still not used Caddy for much yet beyond testing a small website with `Cache-Control` headers. Initially I had issues with the default config from DockerHub in a local environment, and learned to write a small Caddyfile to disable HTTPS. 

Then I later wanted to try HTTPS locally but wasn't sure about the earlier issue where I think it tried to use LetsEncrypt and naturally failed. Found the `local_certs` option which worked (at the time my site address was `:80` and I probably changed to `localhost` at the same time, unaware that also fixed the issue).

Finally, I needed to test a browser behaviour that is meant to only work with HTTPS, and tried to configure for HTTP and HTTPS without the redirect, that's where I got stuck with the docs. The quickstart guide focuses on an actual public server rather than a local dev environment for testing privately on a `localhost` hostname. The intention was to provide some reference/guidance (in lack of quickstart guide) for such a setup, which at the time when I created this PR, I had several misunderstandings of gotchas that I thought was beneficial to inform others about.

> I don't see any correlation to "localhost" sites and it's more about configuring auto HTTPS in the Caddyfile.

New users wanting to use Caddy locally as I've described are probably using Caddyfile config initially(you'd probably know better if that's really the case). The examples all use `localhost` as a site address, and it's about auto HTTPS configs for a reason as both this file and the linked one with examples are about that specific feature.

The snippets are small and may seem obvious to you with your experience in Caddy as the author, but as a new user I stumbled around a lot unsure of what Caddy was wanting me to do configuration wise(eg I didn't realize that `:80` or `localhost:80` would make any difference regarding how certs were handled, I just kept trying different things as I scrambled around the docs trying to make sense of the magic Caddy was doing under the hood and why the behaviour wasn't always seeming consistent(due to the different implicit behaviours based on subtle configuration changes).

So the examples purpose is to save time and confusion for getting what should be fairly simple to setup and support users to [falling into the pit of success](https://blog.codinghorror.com/falling-into-the-pit-of-success/) better.

---

**EDIT:** I've experienced more misunderstandings of what Caddy is doing while refining the example snippets and testing them, I don't want to keep wasting your time, with this so just wait until I've pushed the new changes, I'll then move them to the wiki. Hopefully I get it right this time around :)

src/docs/markdown/automatic-https.md
polarathene: > Why not both?

I'm just not familiar with your wiki and process for that yet. I'm less likely to notice any notifications from other platforms. I'll move the examples to the wiki, and check on it depending how busy I am, otherwise I assume it's less stringent than a PR and the community is able to directly improve it as they see fit.

flaviostutz:
I am reaching this thread by the same reasons as the author (#76 ). I had a bad time at understanding the functional behavior of auto-http "modes". I am not sure if the suggested truth table is correct (I have just tried `auto_https off` and the site is not reachable by plain http as I expected), but something like it would be very helpful. Please fix it and keep a table like that in docs. I, as a user, got very confused on this and the table was the kind of thing I would like to see in docs.

polarathene:
> I, as a user, got very confused on this and the table was the kind of thing I would like to see in docs.

I'm not sure if a table is appropriate, I haven't had time to sort these docs out past few days but will get back to it and refresh on that :+1: 

Yes as a new user to Caddy it's definitely an area that the current docs haven't communicated well. Are you also starting off with using Caddyfile for configuration? I'm to provide a bunch of Caddyfile snippet examples to better explain how it works.

`auto_https` is effectively `https://` prefixed to any site address (except HTTP port), so whatever port you provide will be configured for HTTPS by default (assuming the requirements for activation pass), thus you get implicit HTTPS. 

You then have `auto_https disable_redirects` which removes the implicit HTTP port(default 80) on `http://`, or `auto_https off` which does that plus it removes the automatic TLS handling, `https://` of your site address will not establish a connection successfully, but will still be able to access the port over `http://` which will inform you that it's not over `https://`(which doesn't work anyway...so not sure what the point of that is?). 

`auto_https` never implicitly lets you under any configuration serve content from `http://`, you need to explicitly declare the intent with `http://` prefix to your site address or specifying the HTTP port for it. If you want to serve both HTTPS and HTTP, you would use `auto_https disable_redirects` and provide `https://` and `http://` protocol schemes as prefixes to site address.

flaviostutz:
Great explanation, @polarathene. Thanks! Maybe part of this text should be placed on documentation. Only now it was clear to me that auto_http is not meant to make the site work with plain http, for example.

flaviostutz:
Sharing my "user" experience here...

I was trying to create a Caddyfile that would support both http and https by just "disabling" http with "auto_https". Now I see this is wrong.

The "best" way I found to achieve this is by using "http://" AND/OR "https://" scheme in Caddyfile.

```
http://localhost https://localhost {
	reverse_proxy 10.0.1.8:80
}
```

francislavoie:
Yep, that's exactly the correct approach @flaviostutz :+1:

mholt:
Yeah, that's how it's done in the Caddyfile, at least. The Automatic HTTPS docs can't be specific to just the Caddyfile adapter though.

polarathene:
> Yeah, that's how it's done in the Caddyfile, at least.

While I understand that you can't change this, it does seem odd that `auto_https off` still listens on the HTTPS port/protocol implicitly(or whatever port you provide over `https://`), since you cannot successfully establish a secure connection over `https://`, nor will it allow `http://` (which responds with "Client sent an HTTP request to an HTTPS server."), unless it's the HTTP port or provided `http://` as prefix.

Wouldn't it have been better named `auto_tls off`? I think that's where some common confusion stems from what the name `auto_https off` implies. `auto_https off` I think most would have thought to switch from implicit HTTPS by default to HTTP by default, where TLS for HTTPS could be opt-in per site address instead(useful I think for reverse proxies that terminate HTTPS via Caddy, then proxy to a local service over HTTP(which I don't think has any notable drawbacks at that point and would still be fairly secure?)

> The Automatic HTTPS docs can't be specific to just the Caddyfile adapter though.

I've got plenty of time for finishing this off today, sorry for the delay. I'll try give the related JSON config a go as well, it's only the Caddyfile and JSON that need examples documented for the wiki?

If you'd like to answer prior to me finding out for myself with the JSON config, is the `disable` option exactly equivalent to `auto_https off`? (I remember Francis had shared some JSON config snippets with me earlier somewhere and a CLI command to generate that based off the Caddyfile, so I assume `auto_https off` maps to the JSON `"disable": true`, the three other options regarding certificates/skip, I'm unsure of if those have any equivalent in Caddyfile?

---

Finally one other question, *Automatic HTTPS* is effectively automatic TLS (eg LetsEncrypt with automated renewals) + HTTP->HTTPS redirects, is that all? Is this difficult/complicated to manually configure in Caddyfile or JSON without the feature, or can I have a snippet to share to reveal the "magic" that the feature is providing?

I just want confirmation if it's only meant for HTTPS or whenever there is a need for TLS (where the HTTP->HTTPS redirect can be disabled, but the automatic TLS is still useful for securing connections with TLS that aren't HTTPS servers). AFAIK, [the only difference between HTTP and HTTPS](https://en.wikipedia.org/wiki/HTTPS#Network_layers) is HTTPS uses TLS to secure the connection, but Caddy seems to detect if a request is not via `https://`(at least for HTTP), rather than allow responding with the expected content (which as mentioned is still oddly the case even with `auto_https off`), so I assume that would be an issue for any other protocol that requires TLS. I haven't got around to testing that out to verify yet.

francislavoie:
Ultimately the best source of information is the code. It's pretty well commented, and it's mostly self-contained to this file, so you should be able to figure it out.

https://github.com/caddyserver/caddy/blob/e385be922569c07a0471a6798d4aeaf972facb5b/modules/caddyhttp/autohttps.go

mholt:
Going to close this PR for now then, as it seems there's still a fair amount of work to do.

polarathene:
Updating my branch for this PR doesn't seem to reflect the new commits here. Nothing really left for the branch to provide value now anyway.

I have rewritten the Examples section and posted to the community wiki [here](https://caddy.community/t/making-sense-of-auto-https-and-why-disabling-it-still-serves-https-instead-of-http/9761). It's less formal and mostly Caddyfile focused. I'll leave any revisions/additions to the community. Hopefully it's a helpful document to other new Caddy users :)

bigfoxtail:
> Updating my branch for this PR doesn't seem to reflect the new commits here. Nothing really left for the branch to provide value now anyway.
> 
> I have rewritten the Examples section and posted to the community wiki [here](https://caddy.community/t/making-sense-of-auto-https-and-why-disabling-it-still-serves-https-instead-of-http/9761). It's less formal and mostly Caddyfile focused. I'll leave any revisions/additions to the community. Hopefully it's a helpful document to other new Caddy users :)

Thank you very much, these things helped me a lot, I spent an afternoon for caddy.
The Wiki version is 2.3, but the program version on docker is 2.21, and it took me ages to find this out. Also your article helped me with this problem, which also took me a lot of time.
To be honest, caddy is good, but not as reliable as nginx.
Thank you very much!

francislavoie:
@bigfoxtail the official Docker image requires a manual process involving making a PR against the official-images repo for a maintainer to merge, so it will often be behind. Watch https://github.com/docker-library/official-images/pull/9365 for when it's merged.

> To be honest, caddy is good, but not as reliable as nginx.

That's a weird thing to say on a random issue like this.

bigfoxtail:
> @bigfoxtail the official Docker image requires a manual process involving making a PR against the official-images repo for a maintainer to merge, so it will often be behind. Watch [docker-library/official-images#9365](https://github.com/docker-library/official-images/pull/9365) for when it's merged.
> 
> > To be honest, caddy is good, but not as reliable as nginx.
> 
> That's a weird thing to say on a random issue like this.

Sorry, but it sucks to run into this kind of thing and waste a lot of time.
This basic function google search for a long time can not find a solution, other users submitted PR and was closed.
As a regular user, I don't want to specialize in code, I want to get the program up and running as soon as possible. I just want a reliable manual, so I'm a little disappointed.

mholt:
@bigfoxtail Open source is a collaboration. Unfortunately, complaining won't help things improve.

I think this thread has run its course.

PR #76, opened 2020-08-11
docs: Add link from `auto_https` option to Auto HTTPS doc

flaviostutz:
It took 3h for me to make Caddy work on my machine because when I disabled redirects, I thought Caddy would serve my website using plain http, but this is not true, because Caddy won't event bind port :80 in this case unless you explicitly configure your site to work on port :80.

### DIFF
--- a/src/docs/markdown/caddyfile/options.md
+++ b/src/docs/markdown/caddyfile/options.md
@@ -60,4 +60,4 @@ Possible options are:
 	- **interval** and **burst** allows `<n>` certificate operations within `<duration>` interval.
 - **local_certs** causes all certificates to be issued internally by default, rather than through a (public) ACME CA such as Let's Encrypt. This is useful in development environments.
 - **key_type** specifies the type of key to generate for TLS certificates; only change this if you have a specific need to customize it.
-- **auto_https** configure automatic HTTPS. It can either disable it entirely (`off`) or disable only HTTP-to-HTTPS redirects (`disable_redirects`). Check [Automatic HTTPS](/docs/automatic-https) for details on how these options affects http/https support.
+- **auto_https** configure automatic HTTPS. It can either disable it entirely (`off`) or disable only HTTP-to-HTTPS redirects (`disable_redirects`). See the [Automatic HTTPS](/docs/automatic-https) page for more details.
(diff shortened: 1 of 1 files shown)


src/docs/markdown/caddyfile/options.md
```diff
@@ -60,4 +60,4 @@ Possible options are:
 	- **interval** and **burst** allows `<n>` certificate operations within `<duration>` interval.
 - **local_certs** causes all certificates to be issued internally by default, rather than through a (public) ACME CA such as Let's Encrypt. This is useful in development environments.
 - **key_type** specifies the type of key to generate for TLS certificates; only change this if you have a specific need to customize it.
-- **auto_https** configure automatic HTTPS. It can either disable it entirely (`off`) or disable only HTTP-to-HTTPS redirects (`disable_redirects`).
+- **auto_https** configure automatic HTTPS. It can either disable it entirely (`off`) or disable only HTTP-to-HTTPS redirects (`disable_redirects`). Be aware that if you choose `disable_redirects` or `off`, port :80 won't be open unless a rule explicitelly listens to it on ingress, for example, with `mysite.org:80`.
```
francislavoie: That's not exactly true. Caddy will still use `:80` (or more specifically, the value of the `http_port` global option) for ACME HTTP challenges. 

https://github.com/caddyserver/website/pull/74 should help clarify some of these things.

Rather, I think what could be changed here instead is to link to the Automatic HTTPS documentation page which better describes the behaviours.

flaviostutz: Great discussion on #74... PR #77 sure would have solved my doubts, specially with the following table:

image

flaviostutz:
I reverted the changes made to this branch and just added the link to the automatic-http page.
Is this the right way to do this in this PR?
Thanks.

src/docs/markdown/caddyfile/options.md
```diff
@@ -60,4 +60,4 @@ Possible options are:
 	- **interval** and **burst** allows `<n>` certificate operations within `<duration>` interval.
 - **local_certs** causes all certificates to be issued internally by default, rather than through a (public) ACME CA such as Let's Encrypt. This is useful in development environments.
 - **key_type** specifies the type of key to generate for TLS certificates; only change this if you have a specific need to customize it.
-- **auto_https** configure automatic HTTPS. It can either disable it entirely (`off`) or disable only HTTP-to-HTTPS redirects (`disable_redirects`).
+- **auto_https** configure automatic HTTPS. It can either disable it entirely (`off`) or disable only HTTP-to-HTTPS redirects (`disable_redirects`). Check [Automatic HTTPS](/docs/automatic-https) for details on how these options affects http/https support.
```
francislavoie: I think keeping it simple is fine here.

```suggestion
- **auto_https** configure automatic HTTPS. It can either disable it entirely (`off`) or disable only HTTP-to-HTTPS redirects (`disable_redirects`). See the [Automatic HTTPS](/docs/automatic-https) page for more details.
```

francislavoie (changes requested):

PR #77, opened 2020-08-16
docs: Add headers to reverse_proxy docs for anchor linking

francislavoie:
I commonly want to link people to a certain part of the reverse proxy docs, but since the page is so long and doesn't have headers, it's hard to point to a specific section. Most specifically about the default header behaviour and `X-Forwarded-For` etc. This should help.


mholt (approved):
Cool, good idea. Thank you!

PR #78, opened 2020-08-19
docs: Clarify file imports are relative to current file, more cross-linking

francislavoie:
See https://caddy.community/t/import-space-in-file-paths/9511 for the discussion where this was found to be unclear.


mholt (approved):
This is a good change. Thanks so much!

PR #82, opened 2020-09-05
docs: Add links to Caddyfile directives in code blocks

francislavoie:
This is super cool.

image

Basically, I make a whitelist based on the filenames of the `.md` files in the caddyfile directives directory, then inject that as JSON to `window.CaddyfileDirectives`... then on load in `docs.js`, I transform the `pre.chroma .k` elements, (`.k` is the class that happens to be the one that the chroma lexer spits out for directives, but also subdirectives) if their inner text matches that whitelist to be a link to the relevant page.

The links have no decorations by default (following the base CSS) but I add a `color: inherit;` style on these links so the colour is the same, and not a link colour. So these links are sorta invisible but you can hover on them to find out. Not the most discoverable thing ever, but it's not bad I think.

### DIFF
--- a/src/includes/docs-head.html
+++ b/src/includes/docs-head.html
@@ -1,5 +1,10 @@
 {{include "/includes/head.html"}}
 <link rel="stylesheet" href="/resources/css/docs.css">
 <link rel="stylesheet" href="/resources/css/chroma.css">
+{{$directives := list }}
+{{range $i, $file := (listFiles "/docs/markdown/caddyfile/directives")}}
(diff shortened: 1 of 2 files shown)


mholt (approved):
Wow, this is pretty cool. Let's give it a try! Thanks for this!

francislavoie:
Btw I realized we can do the the same for `[<matcher>]` tokens all over as well. I'll do another PR soon for that. It'll be a separate rule for that, but it should be ezpz.

Could also do the same for `<directives...>` I think, to link to the directives index.

PR #87, opened 2020-09-25
docs: Minor JSON docs bugfix for hiding a header, remove invalid comment

francislavoie:
Nothing fancy here, just some minor fixes :smile: 

Also updated the Caddyfile so that testing the JSON docs and the download page is a bit easier, by proxying to the real site for the underlying data (easier for people who aren't @mholt I guess :smile:)

### DIFF
--- a/src/includes/docs-details.html
+++ b/src/includes/docs-details.html
@@ -10,7 +10,7 @@
 <div id="top-doc">
 	<!--Populated by JS-->
 </div>
-<h2 id="field-list">Field List</h2>
+<h2 id="field-list-header">Field List</h2>
 <dl id="field-list-contents">
 	<!--Populated by JS-->
 </dl>
--- a/src/resources/js/docs.js
+++ b/src/resources/js/docs.js
@@ -50,9 +50,8 @@ $(function() {
 		});
 
 	// Add links to [<matcher>] or named matcher tokens in code blocks.
-	// We can't do it exactly the same as the above block, because
-	// the matcher text includes <> character which are parsed as HTML
-	// unless we use text() to change the link text.
+	// The matcher text includes <> characters which are parsed as HTML,
+	// so we must use text() to change the link text.
 	$('pre.chroma .nd')
 		.map(function(k, item) {
 			let text = item.innerText;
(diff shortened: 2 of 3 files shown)


Caddyfile
```diff
@@ -14,4 +14,6 @@ rewrite /docs/json/*    /docs/json/index.html
 rewrite /docs/modules/* /docs/modules/index.html
 rewrite /docs/*         /docs/index.html
 
-reverse_proxy /api/* localhost:4444
+reverse_proxy /api/* https://caddyserver.com {
+	header_up Host caddyserver.com
+}
```
mholt: Nice hack, but can we not do this? I'm afraid too many people (and 1 is too many) will forget that they're interacting with the live DB.

src/includes/docs-details.html
```diff
@@ -10,7 +10,7 @@ <h2 id="docs">Description</h2>
 <div id="top-doc">
 	<!--Populated by JS-->
 </div>
-<h2 id="field-list">Field List</h2>
+<h2 id="field-list" style="display: none;">Field List</h2>
```
mholt: Let's avoid inline styles -- put it in the .css instead.

src/resources/js/docs.js
```diff
@@ -50,9 +50,6 @@ $(function() {
 		});
 
 	// Add links to [<matcher>] or named matcher tokens in code blocks.
-	// We can't do it exactly the same as the above block, because
-	// the matcher text includes <> character which are parsed as HTML
-	// unless we use text() to change the link text.
```
mholt, line 55: Is this not true anymore?

mholt (changes requested):
Thank you!

src/resources/js/docs.js
francislavoie: It was while I was writing it but then I changed things and force pushed in the last PR and it wasn't true anymore. Only noticed after it was merged. If you look at the 2 blocks, they're basically doing the same thing now.

Caddyfile
francislavoie: Alright

src/includes/docs-details.html
francislavoie: lol turns out there was already a `#field-list-header { display: none; }` but this was using `id="field-list"` :man_facepalming:

src/resources/js/docs.js
francislavoie: Really only the 2nd line of this comment was no longer valid (i.e. referring to "same as the above block") but the note about `text()` was still valid, I put that back in.

mholt (approved):
Cool, thanks -- I had missed the `-header` part.

PR #92, opened 2020-10-08
add Webi installer, and Linux & Windows hints

coolaj86:


francislavoie:
I'm not sure I agree with putting Webi above the deb/rpm ones, those are distro packages we maintain. Right now, you have it in there twice. Not sure that makes sense. I think only the one at the bottom should be left in.

coolaj86:
Well, in fact, you do! Webi downloads the github release.

I’d be happy to add you as a maintainer. I’d also be happy to add a mechanism by which webi pulls the install script from the github repo, though that seems redundant.

I thought about putting it up by static binaries, which is correct, but also awkward.

Where would you like it to go?

coolaj86:
@francislavoie I moved the generic linux section below the mainstream distros section.

src/docs/markdown/download.md
```diff
@@ -10,7 +10,7 @@ All our [official distributions](https://github.com/caddyserver/dist) come with
 
 ## Static binaries
 
-You can **[download Caddy from GitHub](https://github.com/caddyserver/caddy/releases)**, where new releases are immediately published.
+You can **[download Caddy from GitHub](https://github.com/caddyserver/caddy/releases)**, where new releases are immediately published, and place it in your PATH.
```
mholt, line 13: This is probably important; but the below steps don't explain how to do this, and I'm not really sure that's the right thing for our docs to do anyway, since that's basically "how to use your computer" more than "how to use Caddy" -- we can keep this change, but I'm worried it will confuse people when they run the command below and it doesn't appear in their PATH.

src/docs/markdown/download.md
```diff
@@ -58,13 +58,42 @@ RHEL/CentOS 7:
 [**View the Caddy COPR**](https://copr.fedorainfracloud.org/coprs/g/caddy/caddy/)
 
 
+## Linux & Raspberry Pi
```
mholt, line 61: Why do we need to distinguish Linux from Raspberry Pi?

mholt (commented):
Thanks for the PR, AJ! Just a couple questions/thoughts.

src/docs/markdown/download.md
coolaj86: I remember a wise man once said "many of the most common issues come from people not knowing how to use their computer".

I figure that at least then they'll know that there's such thing as a PATH and that caddy should be in it and if they google "how to put caddy in path" the first hit (for me at least) goes to the install docs, here: https://caddyserver.com/docs/install

I don't feel that strongly about it, but I think it's a nice touch.

src/docs/markdown/download.md
coolaj86: I often do in my own documentation for a couple of reasons. It's kind of a "Kleenex" vs "Tissue Paper" king of thing.

Generally speaking "Linux" means not "kernel" but rather "something Debian-ish on AMD64 hardware for web services or general desktop nerdery" whereas "Raspberry Pi" means "Internet of Things, but easier than arduino, on ARM".

Someone searching for "Linux" is a different user than someone searching for "Raspberry Pi".

Also, many modern tutorials for Linux don't readily work on Raspberry Pi because they're based on different sets of tools due to the difficulty of getting everything on an RPi.

I can drop it if you like, but including both terms is a signal to the reader that yes, in fact, Caddy is a product that works on Raspberry Pi (ARM Linux), and Webi in an installer that works there too.

mholt (approved):
Okie. Sounds good to me, let's give it a spin. Thanks!

PR #94, opened 2020-10-09
docs: add x-forwarded-for note to remote_ip documentation

dylanarmstrong:
Just a quick clarifying note that remote_ip will use X-Forwarded-For header if it's included in remote headers.


mholt (approved):
This is a good point. Thanks for the contribution!

PR #100, opened 2020-11-01
docs: Use headers for individual global options, anchor linkable

francislavoie:
- Changes the style of the global options docs to use `<h5>` instead of `<li>`, which makes them anchor linkable. It's a bit more verbose now, but much easier to scan I think.

- Adds some JS to make each option link to the appropriate header on the same page 🎉 

image


mholt (approved):
Let's give it a try! Thanks

PR #101, opened 2020-11-01
docs: Improve the log example to cover hiding IPs

### LINKED ISSUE #3837: Caddyfile ip_mask filter does not mask entries in common_log json section
As per [this post in the forum](https://caddy.community/t/use-ip-mask-in-caddyfile-v2/10173/2) I am using the new `ip_mask` filter for my Caddyfile. This seems to work well (apart from the IPV4 issues described in #3821)

My `logging` section looks like this:
```
(logging) {
    log {
        output file /var/log/access.log
        format filter {
            wrap json
            fields {
                request>remote_addr ip_mask {
                    ipv4 24
                    ipv6 56
                }
            }
        }
    }
}
```

The `json` format that I am `wrap`ing here also contains a `common_log` property though which still contains the non-redacted version of the IP address. See here:
```
{
  "level": "info",
  "ts": 1604248800.1410089,
  "logger": "http.log.access.log2",
  "msg": "handled request",
  "request": {
    "remote_addr": "[2a02:8109:9a80:1500::]:50654",
    "proto": "HTTP/2.0",
    "method": "GET",
    "host": "k-nut.eu",
    "uri": "/",
    "headers": {
      "User-Agent": [
        "curl/7.64.1"
      ],
      "Accept": [
        "*/*"
      ]
    },
    "tls": {
      "resumed": false,
      "version": 771,
      "cipher_suite": 49196,
      "proto": "h2",
      "proto_mutual": true,
      "server_name": "k-nut.eu"
    }
  },
  "common_log": "2a02:8109:9a80:15ab:fca0:b2c:3ab9:96aa - - [01/Nov/2020:17:40:00 +0100] \"GET / HTTP/2.0\" 200 4103",
  "duration": 0.001558001,
  "size": 4103,
  "status": 200,
  "resp_headers": {
    "Server": [
      "Caddy"
    ],
    "Strict-Transport-Security": [
      "max-age=31536000"
    ],
    "Last-Modified": [
      "Sat, 23 May 2020 07:41:07 GMT"
    ],
    "Accept-Ranges": [
      "bytes"
    ],
    "Content-Length": [
      "4103"
    ],
    "X-Content-Type-Options": [
      "nosniff"
    ],
    "X-Frame-Options": [
      "SAMEORIGIN"
    ],
    "Etag": [
      "\"qarxcj35z\""
    ],
    "Content-Type": [
      "text/html; charset=utf-8"
    ]
  }
}
```

So now this puts developers in a weird spot: While the ip_mask technically works, it does not mask the IPs everywhere which seems to somehow defeat the purpose (as it its probably going to be used by developers that want to remove PII as much as possible). 

I therefore think that (preferrably) the filter should be extended to also cover the `common_log` section (if possible) or a section should be added to the documentation that shows how to turn off `common_log` in those cases.

francislavoie:
Closes https://github.com/caddyserver/caddy/issues/3837

Adds a note about `common_log` to the IP masking example.

### DIFF
--- a/src/docs/markdown/caddyfile/directives/log.md
+++ b/src/docs/markdown/caddyfile/directives/log.md
@@ -227,13 +227,14 @@ log {
 }
 ```
 
-Mask the remote address from the request, keeping the first 16 bytes (i.e. 255.255.0.0) for IPv4 addresses, and the first 64 bytes from IPv6 addresses:
+Mask the remote address from the request, keeping the first 16 bytes (i.e. 255.255.0.0) for IPv4 addresses, and the first 64 bytes from IPv6 addresses, and also deletes the `common_log` field which would normally contain an unmasked IP address:
 
 ```caddy-d
 log {
 	format filter {
 		wrap console
 		fields {
+			common_log delete
 			request>remote_addr ip_mask {
 				ipv4 24
 				ipv6 32


mholt (approved):
Good idea, thanks!

PR #103, opened 2020-11-01
docs: Add new proxy transport options, fix `buffer_requests` location

francislavoie:
I think I might need to split this one into two PRs, cause we probably want to get the `buffer_requests` thing merged in sooner rather than later, but the other options will only be released in v2.3.0

I pretty much copied the wording for the two new options from https://golang.org/pkg/net/http/ since it's basically just a pass-through. Hope that's ok.


mholt (approved):
LGTM, thanks. I think it's fine since I probably won't get around to a major site deploy until around 2.3 anyways.

PR #106, opened 2020-11-22
Replaced http.cat example with single line respond

Cubox:
The http.cat example did not work for me, and I found that using a simple respond with {http.error.status_code} {http.error.status_text} gives a good way to make errors handled, without having to delve into templates or making a page for each error code

If you would rather keep the http.cat example and just add mine, sure.
The behaviour from my example could also be implemented by default, instead of showing a blank empty page to the user. But that's a story for another day.

### DIFF
--- a/src/docs/markdown/caddyfile/directives/handle_errors.md
+++ b/src/docs/markdown/caddyfile/directives/handle_errors.md
@@ -53,3 +53,11 @@ handle_errors {
 	}
 }
 ```
+
+Simply use [`respond`](/docs/caddyfile/directives/respond) to return the error code and name
+
+```caddy
+handle_errors {
+	respond "{http.error.status_code} {http.error.status_text}"
+}
+```
--- a/src/docs/markdown/caddyfile/directives/handle_errors.md
+++ b/src/docs/markdown/caddyfile/directives/handle_errors.md
@@ -48,6 +48,8 @@ Reverse proxy to a professional server that is highly qualified for handling HTT
 ```caddy
 handle_errors {
 	rewrite * /{http.error.status_code}
-	reverse_proxy https://http.cat
+	reverse_proxy https://http.cat {
+		header_up Host http.cat
+	}
 }
 ```


francislavoie:
Ah, I didn't realize we had that example. I know the problem with the existing example from debugging it on the forums.

src/docs/markdown/caddyfile/directives/handle_errors.md
```diff
@@ -43,11 +43,10 @@ handle_errors {
...
 ```
 
-Reverse proxy to a professional server that is highly qualified for handling HTTP errors and improving your day:
+Simply use [`respond`](/docs/caddyfile/directives/respond) to return the error code and name
 
 ```caddy
 handle_errors {
-	rewrite * /{http.error.status_code}
-	reverse_proxy https://http.cat
+    respond "{http.error.status_code} {http.error.status_text}"
```
francislavoie: The problem is that `http.cat` expects the Host to match. Adding `header_up` fixes the problem.

```suggestion
	rewrite * /{http.error.status_code}
	reverse_proxy https://http.cat {
		header_up Host {http.reverse_proxy.upstream.hostport}
	}
```

francislavoie: That said, your example is good too, so if you'd like to re-arrange it to fix the `http.cat` bug, and add yours, that'd be great!

mholt:
Darn, that used to work. (I had used it myself which is why I set it as an example.) Guess they changed something on their end.

Will look at this during the week, thanks!

francislavoie:
For ref, we figured out the fix here: 

https://caddy.community/t/error-handling-using-http-cat-and-reverse-proxy-not-functioning-or-im-stupid-one-of-the-two/10485/2

Cubox:
I can change my PR to have both the fix and my example in it.
However, since http.cat changed their behaviour, maybe it's to get us to stop using this example? Are we sure it's OK to keep ~~recommanding~~ *showing how to* redirect all error codes to their website?

mholt:
We're not recommending that, it's just an example.

francislavoie:
Yeah in practice you might use your own backend that you proxy to instead. It's just meant as a fun explanatory example to show what Caddy can do.

Cubox:
I used recommanded by mistake, sorry. Yeah, it's just an example. I'm just wondering if the change they made to requiring Host to be properly set was in reaction to Caddy having it as an example (wanting to stop us) or not.
Fine with me either way

francislavoie:
Doubt it. It's just that their site config expects the host to actually match now, instead of Caddy's default behaviour which is to send the Host as the one from the original request (which would be your own domain).

Cubox:
@francislavoie @mholt I have done the suggested modifications. Fixed up the http.cat example and added mine

src/docs/markdown/caddyfile/directives/handle_errors.md
```diff
@@ -49,5 +49,14 @@ Reverse proxy to a professional server that is highly qualified for handling HTT
 handle_errors {
 	rewrite * /{http.error.status_code}
 	reverse_proxy https://http.cat
+	header_up Host http.cat
```
francislavoie: That's not exactly right - `header_up` is a subdirective of `reverse_proxy`, note the `{ }` in what I gave you. Also using the placeholder generalizes it, in that it uses the hostname from the proxy address (in this case it's always `http.cat` but if someone has more than one upstream, it can be dynamic) and avoids some slight duplication.

francislavoie: There's a similar example in the `reverse_proxy` docs:

https://caddyserver.com/docs/caddyfile/directives/reverse_proxy#examples

Cubox: Oh, good catch. That's my bad

Cubox: Fixed it. Fought with git rebase to make it all pretty

mholt (approved):
Thanks!

PR #107, opened 2020-11-24
docs: Add caddyfile server options, reorganize the page a bit

francislavoie:
Re-ordered some stuff, so hopefully the diff isn't too funky.

### DIFF
--- a/src/docs/markdown/caddyfile/options.md
+++ b/src/docs/markdown/caddyfile/options.md
@@ -20,35 +20,58 @@ Possible options are:
 
 ```caddy
 {
+	# General Options
 	debug
-	http_port   <port>
-	https_port  <port>
-	default_sni <name>
+	http_port  <port>
+	https_port <port>
 	order <dir1> first|last|[before|after <dir2>]
-	experimental_http3
 	storage <module_name> {
 		<options...>
 	}
-	acme_ca <directory_url>
-	acme_ca_root <pem_file>
-	acme_eab <key_id> <mac_key>
-	acme_dns <provider>
-	email   <yours>
 	admin   off|<addr> {
 		origins <origins...>
 		enforce_origin
 	}
+
+	# TLS Options
+	auto_https off|disable_redirects
+	email <yours>
+	default_sni <name>
+	local_certs
+	acme_ca <directory_url>
+	acme_ca_root <pem_file>
+	acme_eab <key_id> <mac_key>
+	acme_dns <provider>
 	on_demand_tls {
 		ask      <endpoint>
 		interval <duration>
 		burst    <n>
 	}
-	local_certs
 	key_type ed25519|p256|p384|rsa2048|rsa4096
-	auto_https off|disable_redirects
 	cert_issuer <name> ...
+
+	# Server Options
+	servers [<listener_address>] {
+		listener_wrappers {
+			<listener_wrappers...>
+		}
+		timeouts {
+			read_body   <duration>
+			read_header <duration>
+			write       <duration>
+			idle        <duration>
+		}
+		max_header_size <size>
+		protocol {
+			allow_h2c
+			experimental_http3
+			strict_sni_host
+		}
+	}
 }
 ```
+
+## General Options
 
 ##### `debug`
 Enables debug mode, which sets all log levels to debug (unless otherwise specified).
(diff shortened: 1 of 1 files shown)


src/docs/markdown/caddyfile/options.md
```diff
@@ -83,27 +123,84 @@ Specifies an External Account Binding to use for all ACME transactions.
...
-Configure automatic HTTPS. It can either disable it entirely (`off`) or disable only HTTP-to-HTTPS redirects (`disable_redirects`). See the [Automatic HTTPS](/docs/automatic-https) page for more details.
-
 ##### `cert_issuer`
 Defines the issuer (or source) of TLS certificates.
+
+
+
+## Server Options
+
+Allows configuring options of the [HTTP servers](/docs/json/apps/http/servers/) that can't otherwise be provided in site blocks.
```
```suggestion
Customizes [HTTP servers](/docs/json/apps/http/servers/) with settings that potentially span multiple sites and thus can't be rightly configured in site blocks. These options affect the listener/socket, or other behavior beneath the HTTP layer.
```

src/docs/markdown/caddyfile/options.md
```diff
@@ -83,27 +123,84 @@ Specifies an External Account Binding to use for all ACME transactions.
...
+- **read_timeout** is a [duration value](/docs/conventions#durations) that sets how long to allow a read from a client's upload. Setting this to a short, non-zero value can mitigate slowloris attacks, but may also affect legitimately slow clients. Defaults to no timeout.
+
+- **read_header_timeout** is a [duration value](/docs/conventions#durations) that sets how long to allow a read from a client's request headers. Defaults to no timeout.
+
+- **write_timeout** is a [duration value](/docs/conventions#durations) that sets how long to allow a write to a client. Note that setting this to a small value when serving large files may negatively affect legitimately slow clients. Defaults to no timeout.
+
+- **idle_timeout** is a [duration value](/docs/conventions#durations) that sets the maximum time to wait for the next request when keep-alives are enabled. Defaults to 5 minutes to help avoid resource exhaustion.
+
+##### `max_header_size`
+The maximum size to parse from a client's HTTP request headers. It accepts all formats supported by [go-humanize](https://github.com/dustin/go-humanize/blob/master/bytes.go).
```
```suggestion
##### `max_header_size`

The maximum size to parse from a client's HTTP request headers. It accepts all formats supported by [go-humanize](https://github.com/dustin/go-humanize/blob/master/bytes.go).
```

src/docs/markdown/caddyfile/options.md
```diff
@@ -83,27 +123,84 @@ Specifies an External Account Binding to use for all ACME transactions.
...
+
+##### `protocol`
+
+- **allow_h2c** enables H2C ("Cleartext HTTP/2" or "H2 over TCP") support, which will serve HTTP/2 over plaintext TCP connections if a client support it. Because this is not implemented by the Go standard library, using H2C is incompatible with most of the other options for this server. Do not enable this only to achieve maximum client compatibility. In practice, very few clients implement H2C, and even fewer require it. This setting applies only to unencrypted HTTP listeners. ⚠️ Experimental feature; subject to change or removal.
+
+- **experimental_http3** enables experimental draft HTTP/3 support. Note that HTTP/3 is not a finished spec and client support is extremely limited. This option will go away in the future. _This option is not subject to compatibility promises._
+
+- **strict_sni_host** require that a request's `Host` header matches the value of the ServerName sent by the client's TLS ClientHello; often a necessary safeguard when using TLS client authentication.
```
mholt: Come to think of it, it's a little weird that this is under protocol, because this doesn't actually configure protocols.

mholt (changes requested):
This is probably a good idea. I like where it's going.

src/docs/markdown/caddyfile/options.md
francislavoie: I was matching the spacing on the ones in other sections, but yeah that's fine

mholt: Hmm are you sure? I thought that's what I was doing.

src/docs/markdown/caddyfile/options.md
francislavoie: I had this argument with @whitestrake months ago when we were hashing out the original plan, and the conclusion we got to was "this is a TLS thing, which is protocol" but yeah it's fuzzy.

mholt: I think this is more of a TLS option, since it does require TLS to make sense.

francislavoie: Sure, but it's still a per-server option though, in JSON. I don't think it's worth making a new `tls` section for this in `servers`, nor do I think it should be an orphan. :man_shrugging:

mholt (approved):

PR #113, opened 2020-11-24
chore: Add .editorconfig

francislavoie:


mholt:
What is this?

francislavoie:
https://editorconfig.org/

Most editors/IDEs (including VSCode) support reading from this file automatically to set defaults for file types

Edit: I guess VSCode doesn't have it by default :man_facepalming: but you can just install this https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig

mholt:
Do any of the maintainers use anything other than VS Code? If we have to opt-in anyway I'm not sure that putting this in the repo does any good.

francislavoie:
It's pretty standard to include it in repos, frankly. It's a declaration of what kind of indentation style you want in that repo. I see no harm in including it, even if it's not enabled for everyone. It'll solve the issue for me, and likely whoever else contributes (who also use an IDE that support it - which is most)

mholt (approved):
Okie, thanks!

PR #114, opened 2020-11-24
docs: Note the HTTP status returned in request body exceeds max_size

pic:


mholt:
Hmm, I am not sure this is correct, or if so, not intentional.

Am going to close this now, but without a proper explanation we cannot accept this.

pic:
I've just tried and 502 is what I've experienced. What would be the expected behaviour?

```
{"level":"error","ts":1606255751.3237627,"logger":"http.log.error.log0","msg":"readfrom tcp omitted:57756->omitted:3000: http: request body too large","request": omitted, "duration":0.003890302,"status":502,"err_id":"h0y5aj0q7","err_trace":"reverseproxy.(*Handler).ServeHTTP (reverseproxy.go:441)"}
```

Highlights: 
**http: request body too large**
**"status":502,"**

mholt:
Thanks; probably 413 then: https://httpstatuses.com/413

Should probably be discussed in an issue for a proposed fix.

francislavoie:
What happens is the `request_body` handler modifies the `r.Body` on the request to throw an error if something tries to read more bytes than the limit.

It just happens that `reverse_proxy` is the next thing to read the body in your config, so it throws that error when attempting to stream the body upstream.

I guess the `reverse_proxy` handler should take care of catching that type of error and returning `413` instead.

FYI we use `http.MaxBytesReader` which is part of stdlib for implementing the limit, so we don't control how the error is returned. We might need to look at the error message contents to determine which status to return... which is annoying, but :man_shrugging: stdlib doesn't give us enums or w/e for the error types.

PR #117, opened 2020-11-26
docs: Mention the auth placeholder in basicauth docs

francislavoie:
https://caddy.community/t/placeholder-for-username-of-basicauth-user/10606

### DIFF
--- a/src/docs/markdown/caddyfile/directives/basicauth.md
+++ b/src/docs/markdown/caddyfile/directives/basicauth.md
@@ -12,6 +12,8 @@ When a user requests a resource that is protected, the browser will prompt the u
 
 Caddy configuration does not accept plaintext passwords; you MUST hash them before putting them into the configuration. The [`caddy hash-password`](/docs/command-line#caddy-hash-password) command can help with this.
 
+After a successful authentication, the `{http.auth.user.id}` placeholder will be available, which contains the authenticated username.
+
 
 ## Syntax
 


src/docs/markdown/caddyfile/directives/basicauth.md
```diff
@@ -12,6 +12,8 @@ When a user requests a resource that is protected, the browser will prompt the u
 
 Caddy configuration does not accept plaintext passwords; you MUST hash them before putting them into the configuration. The [`caddy hash-password`](/docs/command-line#caddy-hash-password) command can help with this.
 
+After a successful authentication, the `{http.auth.user.id}` placeholder will be available, which contains the authenticated username.
```
mholt, line 15:
```suggestion
As with any other authentication backend, after a successful authentication, the `{http.auth.user.id}` placeholder will be available, which contains the authenticated username.
```

mholt (approved):
LGTM, although I have a suggestion since I think it's important for users to understand that this is not limited to just basicauth.

src/docs/markdown/caddyfile/directives/basicauth.md
francislavoie: I think if we word it like that, that'll make users think "what are the other authentication backends?" and they won't find anything useful in the docs. So we should consider how to answer that. We should probably add a section somewhere we can link to.

PR #118, opened 2020-11-29
Added log level values

lensi:


src/docs/markdown/caddyfile/directives/log.md
```diff
@@ -177,7 +177,8 @@ Masks IP addresses in the field using a CIDR mask, i.e. the number of bytes from
 }
 ```
 
-
+### Levels
+The following log levels are available with `INFO` being the default: `DEBUG`, `INFO`, `WARN`, `ERROR`, `PANIC`, and `FATAL`.
```
mholt, line 181: This can go online 106 which is where we document setting the level.

mholt (changes requested):
Thanks for the submission!

src/docs/markdown/caddyfile/directives/log.md
lensi: I don't quite think so... I was trying to add documentation for the top level log level which is only described on line 23, but missing possible values.

Besides, from further working with this it seems only relevant levels are INFO and ERROR? It would be great to have a description of what is INFO and what is ERROR in regards to a request/response.

mholt: What do you mean by relevant? We use all log levels except for PANIC and FATAL (currently).

francislavoie: This specific doc is about access logs only. I don't think we use anything other than INFO and ERROR for access logs.

mholt: I see... but that is probably just coincodence, I don't see why we should restrict ourselves from adding other logs in the future, especially if it's helpful in debugging or warning.

king5300: Fdd

mholt:
Closing due to inactivity

PR #121, opened 2020-12-04
docs: Update url in index.html to fix 404

aikedejongste:
This url gives a 404, can be fixed by removing the /v1/.


francislavoie:
Thanks!

PR #125, opened 2021-01-07
Escaping characters in rewrite directive docs

Klooven:
I escaped `<` and `>` in the markdown as web browsers confuse it for a HTML tag 🙂


mholt (approved):
Ah, thanks! Silly us.

PR #129, opened 2021-01-21
Implements pre-selected packages on download page

BoutValentin:
Adding the possibilities to share link with pre-selected packages and url update when packages is selected or unselected.

Edit: More information :

This pull request implement, when selecting or unselecting a package on the click of the select box, the adding of the URL of the package selected in the website URL by replacing it with the "replaceState" method. This script now also check for query parameters in URL (named "package") to pre-select packages. In that way we can now share link with preselected packages which can be useful for other projects that use caddy to ensure that people choose the good additional packages. The default use for an user would be : He comes on the downloads page and selects additional packages. Now the browser URL contain packages he preselected. He can now copy paste this link and put it in the documentation of his projects. Now when other user will came from this link, they would already have preselected packages and just have to click on the download button.

### DIFF
--- a/src/resources/js/download.js
+++ b/src/resources/js/download.js
@@ -1,12 +1,13 @@
 // download package list as soon as possible
 $.get("/api/packages").done(function(json) {
 	var packageList = json.result;
+	const preselectedPackage = new URL(window.location.href).searchParams.getAll("package") ;
 	
 	// wait until the DOM has finished loading before rendering the results
 	$(function() {
 		const optpkgTemplate =
 			'<tr class="optpkg">'+
-			'	<td><label><input type="checkbox"><span class="optpkg-name"></span></label></td>'+
+			'	<td><label><input type="checkbox" class="optpkg-check"><span class="optpkg-name"></span></label></td>'+
 			'	<td class="text-center"><input type="text" name="version" placeholder="latest" title="Package version" size="5"></td>'+
 			'	<td class="optpkg-modules"></td>'+
 			'</tr>';
@@ -27,6 +28,30 @@ $.get("/api/packages").done(function(json) {
 			
 			var $optpkg = $(optpkgTemplate);
 			$('.optpkg-name', $optpkg).text(pkg.path);
+			if (preselectedPackage.includes(pkg.path)) {
+				$('.optpkg-check', $optpkg).prop("checked", true);
+				$('.optpkg-check', $optpkg).closest('.optpkg').toggleClass("selected");
+			}
+
+			$('.optpkg-check', $optpkg).change({pkg: pkg}, (event) => {
+				const element = $(event.currentTarget);
+				let newUrl = new URL(window.location.href);
+				let currentSelected = newUrl.searchParams.getAll("package") ;
+				newUrl.searchParams.delete("package");
+				const pkgPath = event.data.pkg.path;
+				if (element.is(':checked')) {
+					if (!currentSelected.includes(pkgPath)) {
+						currentSelected = [...currentSelected, pkgPath];
+					}
+				} else {
+					const position = currentSelected.indexOf(pkgPath);
+					if (position >= 0) {
+						currentSelected.splice(position, 1);
+					}
+				}
+				currentSelected.forEach( (selected) => newUrl.searchParams.append("package", selected));
+				history.replaceState({}, "Download Caddy", newUrl.toString());
+			});
 
 			if (pkg.modules && pkg.modules.length > 0) {
 				for (var j = 0; j < pkg.modules.length; j++) {
@@ -43,6 +68,7 @@ $.get("/api/packages").done(function(json) {
 			}
 			$('#optional-packages').append($optpkg);
 		}
+		updatePage();
 	});
 }).fail(function(jqxhr, status, error) {
 	swal({


src/resources/js/download.js
```diff
@@ -27,6 +28,32 @@ $.get("/api/packages").done(function(json) {
 			
 			var $optpkg = $(optpkgTemplate);
 			$('.optpkg-name', $optpkg).text(pkg.path);
+			if (preselectedPackage.includes(pkg.path)) {
+				$('.optpkg-check', $optpkg)[0].checked = true;
```
dunglas:
```suggestion
				$('.optpkg-check', $optpkg).prop("checked", true);
```

src/resources/js/download.js
```diff
@@ -27,6 +28,32 @@ $.get("/api/packages").done(function(json) {
...
+				} else {
+					if (currentSelected.includes(pkgPath)) {
+						const position = currentSelected.indexOf(pkgPath);
+						currentSelected.splice(position, 1);
+					}
+				}
+				const encodeArray = encodeArrayURI(currentSelected);
+				let newUrl = window.location.pathname;
+				if (encodeArray && encodeArray.length > 0) {
+					newUrl = `${window.location.pathname}?package=${encodeArray.join('&package=')}`
```
dunglas: You should use `(new URLSearchParams(window.location.search)).append("package", val)` instead to generate a URL such as `download?package=github.com/foo/bar&package=github.com/foo/baz`.

src/resources/js/download.js
```diff
@@ -27,6 +28,32 @@ $.get("/api/packages").done(function(json) {
...
+				}
+				history.pushState({}, 'Download Caddy', newUrl);
```
dunglas: Adding a new entry in the history is probably not necessary. You can just call `window.location.search = newUrl.search`.

src/resources/js/download.js
```diff
@@ -218,4 +246,16 @@ function enableFields() {
 	window.onbeforeunload = null;
 }
 
-var downloadButtonHtml; // to restore button to its original contents
\ No newline at end of file
+var downloadButtonHtml; // to restore button to its original contents
+
+function getAllPackageFromURLSearch() {
+	return window.location.search ? new URLSearchParams(window.location.search).getAll('package') : [];
```
dunglas:
```suggestion
new URL(window.location.href).searchParams.getAll("package")
```

And so you maybe not need a dedicated function

src/resources/js/download.js
```diff
@@ -218,4 +246,16 @@ function enableFields() {
...
-var downloadButtonHtml; // to restore button to its original contents
\ No newline at end of file
+var downloadButtonHtml; // to restore button to its original contents
+
+function getAllPackageFromURLSearch() {
+	return window.location.search ? new URLSearchParams(window.location.search).getAll('package') : [];	
+}
+
+function getAllPackageFromURLSearchDecode() {
+	return getAllPackageFromURLSearch().map( (link) => decodeURIComponent(link));
```
dunglas: `decodeURIComponent` shouldn't be necessary if you use the `URL` class directly.

src/resources/js/download.js
```diff
@@ -218,4 +246,16 @@ function enableFields() {
...
+}
+
+function encodeArrayURI(toEncode) {
+	return toEncode.map( (url) => encodeURIComponent(url));
```
dunglas: It's not necessary if you use `searchParams.append("package", val)`.

dunglas (changes requested):

src/resources/js/download.js
BoutValentin: Done, Thanks ! :)

src/resources/js/download.js
BoutValentin: Done, Thanks ! :)

src/resources/js/download.js
BoutValentin: I replace it with history.replaceState instead to avoid useless reloading if we touched to window.location.search and useless append in history with pushState :)

src/resources/js/download.js
BoutValentin: Done, Thanks ! :)

src/resources/js/download.js
BoutValentin: Done, Thanks ! :)

src/resources/js/download.js
BoutValentin: Done, Thanks ! :)

mholt:
Thanks; can you briefly describe how it works and how to use it?

src/resources/js/download.js
```diff
@@ -27,6 +28,30 @@ $.get("/api/packages").done(function(json) {
...
+
+			$('.optpkg-check', $optpkg).change({pkg: pkg}, (event) => {
+				const element = $(event.currentTarget);
+				let newUrl = new URL(window.location.href);
+				let currentSelected = newUrl.searchParams.getAll("package") ;
+				newUrl.searchParams.delete("package");
+				const pkgPath = event.data.pkg.path;
+				if (element.is(':checked')) {
+					if (!currentSelected.includes(pkgPath)) {
+						currentSelected = [...currentSelected, pkgPath];
```
francislavoie, line 44: FYI https://caniuse.com/?search=spread%20operator

Not sure what your ideal target for browser support is @mholt but just FYI

mholt: Good find, thanks. That should be okay though; it's also a fairly niche feature. Browsers should be auto-updating these days anyway.

BoutValentin:
> Thanks; can you briefly describe how it works and how to use it?

Yeah of course, I just update The pull request description :)

mholt (approved):
LGTM -- works well too, thanks! My main concern was it cluttering browser history, which it doesn't appear to do.

PR #131, opened 2021-01-27
docs: Shorten the php_fastcgi expanded form a bit

francislavoie:
Just making use of the matcher shortcuts we have available to make it look a little less fluffy.


mholt (approved):
Beautiful.

PR #61, opened 2020-07-03
docs: Clarify Caddyfile address separators

francislavoie:
See https://github.com/caddyserver/caddy/issues/3545

PR #136, opened 2021-02-13
Update concepts.md

basilhendroff:
A follow up to https://github.com/caddyserver/website/pull/61 arising from 
https://caddy.community/t/reverse-proxy-combining-com-and-other-country-tld/11482

### DIFF
--- a/src/docs/markdown/caddyfile/concepts.md
+++ b/src/docs/markdown/caddyfile/concepts.md
@@ -186,6 +186,8 @@ If multiple sites share the same definition, you can list all of them together:
 localhost:8080, example.com, www.example.com
 ```
 
+Notice when addresses are listed on the same line, the comma is optional, however the space between addresses is not, 
+
 or
 
 ```caddy
--- a/src/docs/markdown/caddyfile/concepts.md
+++ b/src/docs/markdown/caddyfile/concepts.md
@@ -186,7 +186,7 @@ If multiple sites share the same definition, you can list all of them together:
 localhost:8080, example.com, www.example.com
 ```
 
-Notice when addresses are listed on the same line, the comma is optional, however, the space between addresses is not, 
+Notice when addresses are listed on the same line, the commas are optional, however, the space between addresses is not, 
 
 or
 
--- a/src/docs/markdown/caddyfile/concepts.md
+++ b/src/docs/markdown/caddyfile/concepts.md
@@ -186,7 +186,7 @@ If multiple sites share the same definition, you can list all of them together:
 localhost:8080, example.com, www.example.com
 ```
 
-Notice when addresses are listed on the same line, the comma is optional, however the space between addresses is not, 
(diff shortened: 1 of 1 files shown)


mholt:
I'm pretty sure we say basically the same thing below: "Notice how the commas indicate the continuation of addresses."

francislavoie:
The trouble @mholt is that it mentions nowhere that commas are optional when on a single line. It just says that commas are required for newlines. Those aren't exactly the same thing.

mholt (changes requested):
I don't think this change makes any sense though. It reads like an incomplete sentence and if I didn't write the parser I don't think I would understand what this means.

How about:

> Spaces are always required between tokens. As a special case, address tokens can wrap to the next line if commas are used to indicate the continuation.

basilhendroff:
Two things:
1. Is the word _token_ necessary? It may scare a novice trying to wrap their head around Caddy so, how about this instead...

> Spaces are always required between addresses. As a special case, addresses can wrap to the next line if commas are used to indicate the continuation.

2. It still doesn't make it obvious that commas are optional if addresses are listed on one line. For instance, this is valid...

`www.domain.com domain.com {` 

...and so is this...

`www.domain.com, domain.com {` 

but this is not...

`www.domain.com# domain.com {` 

The comma is a valid separator, but other special characters are not.

To further reinforce the point, and by way of example, I use a combination of both single-line and multi-line versions in my own Caddyfile to improve readability ...

```
www.caffigoalkeeping.com, caffigoalkeeping.com,
www.caffigoalkeeping.com.au, caffigoalkeeping.com.au {
  ...
  reverse_proxy http://10.1.1.51
}
```
While optional, leaving the first comma on each of the first two lines out, for me, would be grammatically questionable purely from the perspective of English language syntax. In my mind, the optional commas just makes the construct read better.

mholt:
> Spaces are always required between addresses. As a special case, addresses can wrap to the next line if commas are used to indicate the continuation.

This is good with me.

basilhendroff:
@mholt Now @francislavoie and I just have to convince you to say something about the optional comma.

mholt:
I don't think it's really that important; (in fact, it'd be kind of nice if I had enforced commas anyway, but given that the Caddyfile can be used for more than just HTTP servers, I didn't want to enforce a comma for other server types if they didn't make sense there), but if you insist, how about just:

> Spaces are always required between addresses. Addresses can also end with a comma to indicate more, or to wrap to the next line.

basilhendroff:
@mholt I'm glad that we could thrash this one out, but I'm not going to insist on anything. Unless @francislavoie has a strong opinion on this, I'm happy to accept whatever you both finally settle on.

mholt:
Ok, well; I've said my opinions on it, but the current changeset isn't updated, so I'll close this; but if it's updated to what we talked about, we can reopen. Thanks!

PR #139, opened 2021-02-17
docs: Some cleanup, some v2.4.0 additions

francislavoie:
I'm sure I'll have more to add to this in the coming days, but:

- ~~Add `abort` directive docs~~ Matt ninja'd me on this one, but I added a different example
- Add a ~~note~~ similar directives section in `handle` to cross-link to `handle_path` and others
- Add another example in `handle_errors` that shows how an `expression` matcher can be used to pair with it
- Add a cat emoji to `handle_errors` because 😸
- Add `file_server` to one of the `php_fastcgi` examples, it's rare that you'll ever use it without `file_server` so might as well include it there
- Add a TOC to `reverse_proxy` cause it's such a long page. Maybe we'll add one to other pages as well, but TBD
- Clarify the upstream address stuff a bit after some discussion in https://caddy.community/t/reverse-proxy-with-multiple-different-upstreams-with-paths/11512/12 and mention `rewrite` as the alternative
- Use the `{re.*.*}` shortcut in the Caddyfile matcher docs, links to the placeholder mapping to let people trace where that comes from

Round 2:
- Bump the minimum Go version to 1.15 in the build page
- Add the `tls` `resolvers` Caddyfile docs
- Add the optional `<directory_url>` on the `acme` issuer
- Add the proxy `health_headers` active health check docs
- Add the proxy `buffer_responses` and `max_buffer_size` docs
- Add `ocsp_stapling` global option
- Mention `cert_issuers` global may be repeated
- Add `--force` option for the `caddy reload` command

### DIFF
--- a/src/docs/markdown/caddyfile/directives/abort.md
+++ b/src/docs/markdown/caddyfile/directives/abort.md
@@ -15,8 +15,18 @@ abort [<matcher>]
 
 ## Examples
 
-Abort all requests for paths starting with `/foo`:
+Forcefully close a connection received for unknown domains when using a wildcard certificate:
 
-```caddy-d
-abort /foo*
+```caddy
+*.example.com {
+    @foo host foo.example.com
+    handle @foo {
+        respond "This is foo!" 200
+    }
+
+    # Unhandled domains fall through to here, but we don't want to accept their requests
+    handle {
+        abort
+    }
+}
 ```
--- a/src/docs/markdown/caddyfile/directives/handle.md
+++ b/src/docs/markdown/caddyfile/directives/handle.md
@@ -8,6 +8,10 @@ Evaluates a group of directives mutually exclusively from other `handle` blocks 
 
 The `handle` directive is kind of similar to the `location` directive from nginx config: the first matching `handle` block will be evaluated. Handle blocks can be nested if needed. Only HTTP handler directives can be used inside handle blocks.
 
+<aside class="tip">
+	See also the [`handle_path`](/docs/caddyfile/directives/handle_path) directive if you need to strip the matched path prefix.
+</aside>
+
 ## Syntax
 
 ```caddy-d
--- a/src/docs/markdown/caddyfile/directives/handle.md
+++ b/src/docs/markdown/caddyfile/directives/handle.md
@@ -8,10 +8,6 @@ Evaluates a group of directives mutually exclusively from other `handle` blocks 
 
 The `handle` directive is kind of similar to the `location` directive from nginx config: the first matching `handle` block will be evaluated. Handle blocks can be nested if needed. Only HTTP handler directives can be used inside handle blocks.
 
-<aside class="tip">
-	See also the [`handle_path`](/docs/caddyfile/directives/handle_path) directive if you need to strip the matched path prefix.
-</aside>
-
 ## Syntax
 
 ```caddy-d
@@ -27,6 +23,14 @@ handle [<matcher>] {
 
 If you prefer crafting HTTP handler logic in a more inheritence-based way like nginx location blocks, you may prefer the use of `handle` blocks rather than defining mutually-exclusive matchers for your directives. If inheritence is a desired characteristic of your HTTP handler configurations, then the `handle` directive may suit you well.
 
+## Similar directives
+
+There are other directives that can wrap other HTTP handler directives, but but imply a different intent or behave subtly differently:
+
+- [`handle_path`](handle_path) does the same as `handle`, but also has built-in path prefix stripping logic built-in before handling the directives within.
+- [`handle_errors`](handle_errors), similarly to `handle`, wraps other directives, but is instead invoked when Caddy encounters an error during request handling.
+- [`route`](route) similarly to `handle`, wraps other directives, but instead of providing mutual exclusivity, allows overriding the default [directive order](/docs/caddyfile/directives#directive-order).
+
 ## Examples
 
 Handle requests in `/foo/` by the static file server, and send all other requests to the reverse proxy:
@@ -37,5 +41,17 @@ handle /foo/* {
 }
 handle {
 	reverse_proxy 127.0.0.1:8080
+}
+```
+
+You can mix `handle` and [`handle_path`](handle_path) directives in the same site, and they will still be mutually exclusive from eachother:
+
+```caddy-d
+handle_path /foo/* {
+	# The path has the "/foo" prefix stripped
+}
+
+handle /bar/* {
+	# The path still retains "/bar"
 }
 ```
--- a/src/docs/markdown/caddyfile/directives/handle.md
+++ b/src/docs/markdown/caddyfile/directives/handle.md
@@ -25,11 +25,11 @@ If you prefer crafting HTTP handler logic in a more inheritence-based way like n
 
 ## Similar directives
 
-There are other directives that can wrap other HTTP handler directives, but but imply a different intent or behave subtly differently:
+There are other directives that can wrap HTTP handler directives, but each has its use depending on the behavior you want to convey:
 
-- [`handle_path`](handle_path) does the same as `handle`, but also has built-in path prefix stripping logic built-in before handling the directives within.
-- [`handle_errors`](handle_errors), similarly to `handle`, wraps other directives, but is instead invoked when Caddy encounters an error during request handling.
-- [`route`](route) similarly to `handle`, wraps other directives, but instead of providing mutual exclusivity, allows overriding the default [directive order](/docs/caddyfile/directives#directive-order).
+- [`handle_path`](handle_path) does the same as `handle`, but it strips a prefix from the request before running its handlers.
+- [`handle_errors`](handle_errors) is like `handle`, but is only invoked when Caddy encounters an error during request handling.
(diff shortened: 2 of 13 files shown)


src/docs/markdown/caddyfile/directives/reverse_proxy.md
```diff
@@ -6,6 +6,17 @@ title: reverse_proxy (Caddyfile directive)
...
+- [Upstreams](#upstreams)
+- [Load balancing](#load-balancing)
+  - [Active health checks](#active-health-checks)
+  - [Passive health checks](#passive-health-checks)
+- [Streaming](#streaming)
+- [Headers](#headers)
+- [Transports](#transports)
+  - [The `http` transport](#the-http-transport)
+  - [The `fastcgi` tranport](#the-fastcgi-transport)
+- [Examples](#examples)
```
mohammed90, line 19: Is it worth the investment of writing ToC extension for goldmark in Caddy? Hugo has done something similar

francislavoie: :man_shrugging: I don't think these are hard to write, and if we need them we can type them up in a minute or two. Also not that hard to maintain, these aren't high-churn pages. But I mean, if you feel like writing code, I won't stop you :joy:

src/docs/markdown/caddyfile/directives/handle.md
```diff
@@ -8,6 +8,10 @@ Evaluates a group of directives mutually exclusively from other `handle` blocks
 
 The `handle` directive is kind of similar to the `location` directive from nginx config: the first matching `handle` block will be evaluated. Handle blocks can be nested if needed. Only HTTP handler directives can be used inside handle blocks.
 
+<aside class="tip">
```
mholt: This is fine, but what about something like "Similar directives"? https://caddyserver.com/docs/caddyfile/directives/rewrite#similar-directives

src/docs/markdown/caddyfile/directives/abort.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Syntax
+
+```caddy-d
+abort [<matcher>]
+```
+
+## Examples
+
+Forcefully close a connection received for unknown domains when using a wildcard certificate:
```
mholt, line 18: This is a good example, better than the one I pushed. Would be OK keeping this. Make sure it's not `caddy-d` though since it's a whole Caddyfile.

src/docs/markdown/caddyfile/directives/reverse_proxy.md
mholt: I don't have much of a preference. We have some room on the right side of the page if we want to ToC every page. But yeah, it does seem like a lot of work to set that up.

mholt (changes requested):
Thanks, these are all good improvements. Sorry my work from this weekend stepped on yours. I had something come up and couldn't commit right away.

francislavoie:
Alright, I added a bunch more stuff, and addressed the review comments.

Might add the `handle_response` Caddyfile stuff that's still pending review if that gets merged before this does, but otherwise this is ready to go I think

src/docs/markdown/caddyfile/directives/handle.md
```diff
@@ -23,6 +23,14 @@ handle [<matcher>] {
 
 If you prefer crafting HTTP handler logic in a more inheritence-based way like nginx location blocks, you may prefer the use of `handle` blocks rather than defining mutually-exclusive matchers for your directives. If inheritence is a desired characteristic of your HTTP handler configurations, then the `handle` directive may suit you well.
 
+## Similar directives
+
+There are other directives that can wrap other HTTP handler directives, but but imply a different intent or behave subtly differently:
```
mholt: This section could use improvement, including in the other linked pages where this section exists. I'll just comment on this one.

```suggestion
There are other directives that can wrap HTTP handler directives, but each has its use depending on the behavior you want to convey:
```

src/docs/markdown/caddyfile/directives/handle.md
```diff
@@ -23,6 +23,14 @@ handle [<matcher>] {
...
+
+- [`handle_path`](handle_path) does the same as `handle`, but also has built-in path prefix stripping logic built-in before handling the directives within.
```
mholt:
```suggestion
- [`handle_path`](handle_path) does the same as `handle`, but it strips a prefix from the request before running its handlers.
```

Just trying to make it a little easier to understand/read.

src/docs/markdown/caddyfile/directives/handle.md
```diff
@@ -23,6 +23,14 @@ handle [<matcher>] {
...
+- [`handle_errors`](handle_errors), similarly to `handle`, wraps other directives, but is instead invoked when Caddy encounters an error during request handling.
```
mholt:
```suggestion
- [`handle_errors`](handle_errors) is like `handle`, but is only invoiked when Caddy encounters an error during request handling.
```

src/docs/markdown/caddyfile/directives/handle.md
```diff
@@ -23,6 +23,14 @@ handle [<matcher>] {
...
+- [`route`](route) similarly to `handle`, wraps other directives, but instead of providing mutual exclusivity, allows overriding the default [directive order](/docs/caddyfile/directives#directive-order).
```
mholt:
```suggestion
- [`route`](route) wraps other directives like `handle` does, but with two distinctions: 1) route blocks are not mutually exclusive to each other, and 2) directives within a route are not [re-ordered]([directive order](/docs/caddyfile/directives#directive-order), giving you more control if needed.
```

src/docs/markdown/caddyfile/directives/handle.md
```diff
@@ -35,3 +43,15 @@ handle {
 	reverse_proxy 127.0.0.1:8080
 }
 ```
+
+You can mix `handle` and [`handle_path`](handle_path) directives in the same site, and they will still be mutually exclusive from eachother:
```
mholt:
```suggestion
You can mix `handle` and [`handle_path`](handle_path) directives in the same site, and they will still be mutually exclusive from each other:
```

src/docs/markdown/caddyfile/directives/reverse_proxy.md
```diff
@@ -67,7 +87,13 @@ Upstream addresses can take the form of a conventional [Caddy network address](/
...
+Note: Schemes cannot be mixed, since they modify the common transport configuration (a TLS-enabled transport cannot carry both HTTPS and plaintext HTTP). Specifying ports 80 and 443 are the same as specifying the HTTP and HTTPS schemes, respectively. Any explicit transport configuration will not be overwritten, and omitting schemes or using other ports will not assume a particular transport.
+
+Additionally, upstream addresses cannot contain paths or query strings, as that would imply simultaneous rewriting the request while proxying, which behavior is not defined or supported. You may use the [`rewrite`](/docs/caddyfile/directives/rewrite) directive should you need this.
+
+If the address is not a URL (i.e. does not have a scheme), then placeholders can be used, but this makes the upstream dynamic.
```
mholt:
```suggestion
If the address is not a URL (i.e. does not have a scheme), then placeholders can be used, but this makes the upstream dynamic, meaning that the potentially many different backends act as one upstream in terms of health checks and load balancing.
```

src/docs/markdown/command-line.md
```diff
@@ -234,6 +235,8 @@ Because this command uses the API, the admin endpoint must not be disabled.
 
 `--address` needs to be used if the admin endpoint is not listening on the default address and if it is different from the address in the provided config file. Note that only TCP addresses are supported at this time.
 
+`--force` will cause a reload to happen even if the specified config is the same as the one the running Caddy instance is already using. Can be useful to force Caddy to reload TLS certificate files specified to load in the config, if they've become stale.
```
mholt:
```suggestion
`--force` will cause a reload to happen even if the specified config is the same as what Caddy is already running. Can be useful to force Caddy to reprovision its modules, which can have side-effects, for example: reloading manually-loaded TLS certificates.
```

mholt (changes requested):
Thanks Francis! Just a quick first pass but I think this'll catch most of the things.

src/docs/markdown/caddyfile/directives/handle.md
francislavoie: :+1: I was somewhat copying the wording from the `rewrite` page.

src/docs/markdown/caddyfile/directives/abort.md
```diff
@@ -15,8 +15,18 @@ abort [<matcher>]
...
+*.example.com {
+    @foo host foo.example.com
+    handle @foo {
+        respond "This is foo!" 200
+    }
+
+    # Unhandled domains fall through to here, but we don't want to accept their requests
+    handle {
+        abort
+    }
```
mholt, line 30: I'm pretty sure the `handle` block isn't needed for this right?

francislavoie: Probably not but it's a mutual exclusivity pattern :man_shrugging:

mholt (approved):
These look good. Thanks Francis!

PR #141, opened 2021-02-20
add custom binary support files docs

caquino:
Hi,

This PR adds initial documentation of the procedure to utilize package support files with custom builds.

This will allow users to use systemd services, bash-completion and default configuration from `caddy` package while using custom binaries.

For more background information please check [this issue](https://github.com/caddyserver/dist/issues/54)

### DIFF
--- a/src/docs/markdown/build.md
+++ b/src/docs/markdown/build.md
@@ -84,7 +84,7 @@ Procedure:
 </code></pre>
 
 
-`dpkg-divert` will move `/usr/bin/caddy` binary to `/usr/bin/caddy.default` and put a divertion in place in case any package want to install a file to this location.
+`dpkg-divert` will move `/usr/bin/caddy` binary to `/usr/bin/caddy.default` and put a diversion in place in case any package want to install a file to this location.
 
 `update-alternatives` will create a symlink from the desired caddy binary to `/usr/bin/caddy`
 
--- a/src/docs/markdown/build.md
+++ b/src/docs/markdown/build.md
@@ -64,3 +64,30 @@ Or similarly for Linux ARMv6 when you're not on Linux or on ARMv6:
 The same works for `xcaddy`. To cross-compile for macOS:
 
 <pre><code class="cmd bash">GOOS=darwin xcaddy build</code></pre>
+
+## Package support files for custom builds for Debian/Ubuntu/Raspbian
+
+This procedure aims to simplify running custom `caddy` binaries while keeping support files from the `caddy` package.
+
+This procedure allows users to take advantage of the default configuration, systemd service files and bash-completion from the official package.
+
+Requirements:
+- Install `caddy` package according to these [instructions](https://caddyserver.com/docs/install#debian-ubuntu-raspbian)
+- Build your custom `caddy` binary according to the procedure listed in this document. (see above)
+- Your custom `caddy` binary should be located in the current directory.
+
+Procedure:
+<pre><code class="cmd"><span class="bash">dpkg-divert --divert /usr/bin/caddy.default --rename /usr/bin/caddy</span>
+<span class="bash">mv ./caddy /usr/bin/caddy.custom</span>
+<span class="bash">update-alternatives --install /usr/bin/caddy caddy /usr/bin/caddy.default 10</span>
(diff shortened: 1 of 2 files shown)


src/docs/markdown/build.md
```diff
@@ -64,3 +64,30 @@ Or similarly for Linux ARMv6 when you're not on Linux or on ARMv6:
...
+
+Procedure:
+<pre><code class="cmd"><span class="bash">dpkg-divert --divert /usr/bin/caddy.default --rename /usr/bin/caddy</span>
+<span class="bash">mv ./caddy /usr/bin/caddy.custom</span>
+<span class="bash">update-alternatives --install /usr/bin/caddy caddy /usr/bin/caddy.default 10</span>
+<span class="bash">update-alternatives --install /usr/bin/caddy caddy /usr/bin/caddy.custom 50</span>
+</code></pre>
+
+
+`dpkg-divert` will move `/usr/bin/caddy` binary to `/usr/bin/caddy.default` and put a divertion in place in case any package want to install a file to this location.
```
francislavoie:
```suggestion
`dpkg-divert` will move `/usr/bin/caddy` binary to `/usr/bin/caddy.default` and put a diversion in place in case any package want to install a file to this location.
```

francislavoie:
I think we should also add a note in https://github.com/caddyserver/website/blob/master/src/docs/markdown/install.md#debian-ubuntu-raspbian linking to this section, like "If you wish to run a custom build of Caddy, follow these instructions ..."

francislavoie:
Tested this out on a server of mine and it worked perfectly 🎉 

FYI @Mohammed90

francislavoie (approved):

mholt (changes requested):
Thanks for this. I'm wondering if this is a better fit on our wiki rather than on our docs (do we want to officially support this method of installation?)...

In any case, I am pretty certain that the instructions should be on the install page, not build.md, because the instructions are more about installing a custom build than actually yielding a new build; and it's very platform-specific. So move it into install.md and then I'll give it another review. Thanks!

francislavoie:
> do we want to officially support this method of installation?

I think we do. We provide `caddy` and `xcaddy` as a .deb package, and it's not uncommon for people to want plugins (obviously). Status quo is broken because if people do `apt upgrade` after we publish a new version, their custom build of Caddy will get wiped out, if they had just replaced the binary in `/usr/bin/caddy`. That's no good.

> In any case, I am pretty certain that the instructions should be on the install page, not build.md

I think keeping `install.md` focused as the "here's how you install vanilla Caddy" page is better. These instructions are only relevant if you actually need a custom build, which is why it's most relevant in `build.md`.

mholt:
> I think keeping install.md focused as the "here's how you install vanilla Caddy" page is better. These instructions are only relevant if you actually need a custom build, which is why it's most relevant in build.md.

I'm gonna hard disagree on that one. These instructions have nothing to do with building Caddy, only with integrating with a package manager. If we're going to accept this into the official docs, it'll have to go in the install page, otherwise it can go in our wiki.

francislavoie:
> These instructions have nothing to do with building Caddy, only with integrating with a package manager.

It has to do with _maintaining_ a build of Caddy. That's very important to document otherwise people will shoot themselves in the foot; it's already happened, and this is what this aims to resolve.

This stuff has to do with pairing the use of `xcaddy` with package managers. The `install.md` docs don't have anything to do with `xcaddy`. Adding stuff about that there will only bloat those instructions, and we want to keep those as streamlined as possible for people to quickly get started. The `build.md` page is for people who dig deeper and need more out of Caddy, and _that's_ when they would need these instructions.

Think of it from a user's perspective. If these instructions were on the `install.md` page, then they would read the cloudsmith install instructions, finish all that, then see they need to build with `xcaddy` so they would go to the `build.md` page, then after they're done building they'll need to go back to the `install.sh` page to do the wiring to maintain the custom build. Having to bounce between the pages like that doesn't quite make sense to me. With the instructions on `build.md`, then they first hit `install.md`, do that, then follow the link to `build.md` to learn how to make their custom build _and_ maintain it. Better UX.

francislavoie:
Could we discuss this again @mholt? I feel pretty strongly that it should be on the `/build` page.

mholt (approved):
Okay,  let's give this a shot.

PR #142, opened 2021-03-14
docs: Add wildcard certificates common pattern

francislavoie:
We did recently add a similar example to the `abort` directive docs, but I feel this is a good fit in this page as well since it's a commonly repeated pattern we respond with when answering forum questions.


mholt (approved):
This is a good idea. My only nit would be that the DNS challenge requirement is currently CA-enforced, so I don't think ZeroSSL has that requirement. (But it might become required for all CAs later, will have to wait and see.)

PR #145, opened 2021-03-14
docs: Add `map` regexp capture groups

francislavoie:
https://github.com/caddyserver/caddy/pull/3991

The `outputs` bit was getting a bit long so I split it up into multiple paragraphs.

The capture group explanation is copied from the `path_regexp` matcher docs.

Added bullets in front of the example cases, because the newlines get collapsed in markdown otherwise, unless we either add bullets or double newlines. I think bullets work better here.


mholt (approved):
Great, thanks for these improvements Francis.

PR #149, opened 2021-03-31
Fix default behavior of field in header directive

clement-escolano:
Hello,

I had the same issue as https://github.com/caddyserver/caddy/issues/3665
I think the documentation of the `field` should be changed to clarify the default behaviour.

Cheers

### DIFF
--- a/src/docs/markdown/caddyfile/directives/header.md
+++ b/src/docs/markdown/caddyfile/directives/header.md
@@ -21,7 +21,7 @@ header [<matcher>] [[+|-|?]<field> [<value>|<find>] [<replace>]] {
 }
 ```
 
-- **&lt;field&gt;** is the name of the header field. By default, will overwrite any existing field of the same name. Prefix with `+` to add the field instead of replace, or prefix with `-` to remove the field.
+- **&lt;field&gt;** is the name of the header field. By default, will add a new field. If `defer` is enabled, it will overwrite any existing field with the same name. Prefix with `+` to always add the field, or prefix with `-` to remove the field.
 - **&lt;value&gt;** is the header field value, if adding or setting a field.
 - **&lt;default_value&gt;** is the header field value that will be set only if the header does not already exist.
 - **&lt;find&gt;** is the substring or regular expression to search for.


mholt:
I don't think this change is correct; without + or - prefix, the default is to set the header to the given value, not append (add) or delete a field. The linked issue is about using with proxied headers, which get manipulated later in the chain.

Will close this, but feel free to continue discussion if I've misunderstood something.

clement-escolano:
I was misleaded (and I think it is the same for the person in the issue) by the fact that setting a header via the `header` directive which is already set by upstream when reverse proxying actually append the header and does not replace it.
This behaviour is not described in the documentation and this is what I am trying to do here.

Do you agree with the behaviour I am describing? If so, what would be the way you want to write it in the documentation?

mholt:
Oh, I see. What you're experiencing is Caddy's `reverse_proxy` handler appending headers from the backend to the downstream response, rather than overwriting them. I guess our docs don't really clarify what it does with _response_ headers, but the basic answer is that it non-destructively appends them rather than replaces existing ones.

clement-escolano:
You found my problem and that's what I was trying to clarify here.
For some reason, updating the branch in my fork does not update the diff here. Is this new wording OK for you? I appended (pun intended) the two last sentences.

---

- **&lt;field&gt;** is the name of the header field. By default, will overwrite any existing field of the same name. Prefix with `+` to add the field instead of replace, or prefix with `-` to remove the field. When using `reverse_proxy`, the default behaviour is to append the header instead of overwriting existing ones. You may want to use `defer` in this case.

francislavoie:
I don't think that's quite right, because the default is always to "set" regardless of whether `reverse_proxy` is used; it's the `reverse_proxy` handler which is appending headers.

Basically you were seeing:

- request begins to be handled
- `header` sets a response header
- `reverse_proxy` does a round trip, then appends the same header
- (should you set `defer`, then) `header` sets a response header

The documentation that should be updated is the `reverse_proxy`'s, not `header`'s, probably.

And you could probably add a note in the `header` docs that describes the interaction with `reverse_proxy`, but probably not inside the `<field>` part cause that's conflating things (this behaviour doesn't really have anything to do with `<field>`, really)

mholt:
> The documentation that should be updated is the reverse_proxy's, not header's, probably.

Yep, I think this is exactly what needs to happen, we should improve the proxy's docs for how it handles response headers.

PR #150, opened 2021-04-04
docs: mention caddy on Arch Linux

shibumi:


mholt (changes requested):
LGTM, we just need a link in the respective place in the menu at the top. :+1: Thanks!

shibumi:
@mholt  fixed. Does this work with the link to "arch-linux-manjaro-parabola"? I am not sure if this gets correctly translated to "Arch Linux, Manjaro, Parabola" note the space between "Arch Linux".

mholt (approved):
Yep, that should work fine! Thanks so much.

PR #151, opened 2021-04-05
docs: Add acme-servers.md and update docs-nav.html

sulliops:
Added a new page with a list of public ACME endpoints, which mirrors the old docs.HTTPS.dev page with similar content. This may be helpful for new Caddy users wanting to make use of Caddy 2.3's multiple certificate issuers feature (see: https://github.com/caddyserver/caddy/pull/3862). Updated docs-nav.html accordingly.

### DIFF
--- a/src/docs/markdown/acme-servers.md
+++ b/src/docs/markdown/acme-servers.md
@@ -1,0 +1,46 @@
+---
+title: "List of ACME Servers"
+---
+
+# List of ACME Servers
+
+**All known, public ACME servers.**
+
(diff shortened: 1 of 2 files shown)


mholt:
Thanks for this contribution -- unfortunately I had to shut down https.dev because the domain name cost over $700 a year. (I was hoping that was just an initial reg fee, but NOPE).

While I think it is good to have this info, it's not really about Caddy, so our docs are the wrong place for them. Maybe our [community wiki](https://caddy.community/c/wiki/13), but we could just as well link to the existing document from https.dev: https://github.com/https-dev/docs/blob/master/list-of-acme-servers.md

(Going to close, thanks for understanding!)

PR #154, opened 2021-04-24
docs: Add basicauth example for when pairing with `handle_errors`

francislavoie:
Context: https://caddy.community/t/problem-with-basicauth-handle-errors/12243

This should give us something to point to when people run into problems with `basicauth` when having `handle_errors` defined.


francislavoie:
Turns out this is incorrect. The behaviour is actually buggy and not working as intended in certain situations https://caddy.community/t/problem-with-basicauth-handle-errors/12243/9

PR #157, opened 2021-04-24
docs: Add `/reverse_proxy/upstreams` API docs

francislavoie:
See https://github.com/caddyserver/caddy/pull/4125


src/docs/markdown/api.md
```diff
@@ -228,3 +230,26 @@ but with an ID, the path becomes
 ```
 
 which is much easier to remember and write by hand.
+
+
+## GET /reverse_proxy/upstreams
+
+Returns the current status of the configured reverse proxy upstreams. Returns a JSON body.
```
mholt:
```suggestion
Returns the current status of the configured reverse proxy upstreams in a JSON document.
```

mholt (commented):
LGTM, just a nit on some awkward phrasing.

It might be worth making a more concise distinction between "healthy" and "available" -- the last paragraph kind of tries to explain this, but instead does so from the angle of active and passive health checks. As we discussed, whether a host is "available" is a property of the actual RP handler, not the host itself, and that's what the paragraph rightly explains, but I think it could be clearer. I will try to write up a suggestion if I have time...

Thanks for completing this!

francislavoie:
Sure 👍 I'll let you re-review this when you can. You often find better wording, but at least this gives you the basis for this content 😄

src/docs/markdown/api.md
```diff
@@ -228,3 +230,26 @@ but with an ID, the path becomes
 ```
 
 which is much easier to remember and write by hand.
+
+
+## GET /reverse_proxy/upstreams
+
+Returns the current status of the configured reverse proxy upstreams in a JSON document.
```
mholt:
```suggestion
Returns the current status of the configured reverse proxy upstreams (backends) as a JSON document.
```

src/docs/markdown/api.md
```diff
@@ -228,3 +230,26 @@ but with an ID, the path becomes
...
+[
+	{"address": "10.0.1.1:80", "healthy": true, "num_requests": 4, "fails": 2},
+	{"address": "10.0.1.2:80", "healthy": true, "num_requests": 5, "fails": 4},
+	{"address": "10.0.1.3:80", "healthy": true, "num_requests": 3, "fails": 3}
+]</code></pre>
+
+Each entry in the JSON array is a configured [upstream](/docs/json/apps/http/servers/routes/handle/reverse_proxy/upstreams/) stored in the global upstream pool.
+
+- **address** is the dial address of the upstream. For SRV upstreams, this is the `lookup_srv` DNS name.
+- **healthy** is the status of whether Caddy knows the upstream to be healthy or not, based on the result of active health checks.
```
mholt:
```suggestion
- **healthy** reflects whether Caddy believes the upstream to be healthy or not. Note that "health" is a distinct concept from "availability". A healthy backend might not be available to proxy to, but an unhealthy backend will always be unavailable. Health is a global characteristic regardless of specific reverse proxy handler configuration, whereas availability is determined by the configuration of the specific reverse proxy handler. For example, a healthy backend would be unavailable if the handler is configured to only allow N requests at a time and it currently has N active requests. The "healthy" property does not reflect availability.
```

src/docs/markdown/api.md
```diff
@@ -228,3 +230,26 @@ but with an ID, the path becomes
...
+- **num_requests** is the amount of active requests currently being handled by the upstream.
+- **fails** the amount of failed requests that are remembered from passive health checks.
```
mholt:
```suggestion
- **fails** the current number of failed requests remembered, as configured by passive health checks.
```

src/docs/markdown/api.md
```diff
@@ -228,3 +230,26 @@ but with an ID, the path becomes
...
+
+The `healthy` status only reflects the result of [active health checks](/docs/json/apps/http/servers/routes/handle/reverse_proxy/health_checks/active/) that have been performed against the backend.
+
+If you've enabled [passive health checks](/docs/json/apps/http/servers/routes/handle/reverse_proxy/health_checks/passive/) for your proxies, then you need to also take into consideration the `fails` and `num_requests` amounts to determine if an upstream is considered available. For accuracy, you should check that the `fails` amount is less than your configured maximum amount of failures for your proxy (i.e. [`max_fails`](/docs/json/apps/http/servers/routes/handle/reverse_proxy/health_checks/passive/max_fails/)), and that `num_requests` is less than or equal to your configured amount of maximum requests per upsteam (i.e. [`unhealthy_request_count`](/docs/json/apps/http/servers/routes/handle/reverse_proxy/health_checks/passive/unhealthy_request_count/) for the whole proxy, or [`max_requests`](/docs/json/apps/http/servers/routes/handle/reverse_proxy/upstreams/max_requests/) for individual upstreams).
```
mholt: So I read the code, and you're right, currently this is true; but I think that's just by accident. I probably didn't think to call `SetHealthy(false)` inside the passive health checks for whatever reason. But there might not be a reason we can't. Perhaps to be conservative, I chose not to, since a slow request might not indicate global unhealthiness, or something like that.

I'd like to be careful that we don't codify an explicit distinction in our docs that `health = active health checks` and `availability = passive health checks`, just in case that changes later.

```suggestion
If your goal is to determine a backend's _availability_, you will need to cross-check relevant properties of the upstream against the handler configuration you are utilizing. For example, if you've enabled [passive health checks](/docs/json/apps/http/servers/routes/handle/reverse_proxy/health_checks/passive/) for your proxies, then you need to also take into consideration the `fails` and `num_requests` values to determine if an upstream is considered available: check that the `fails` amount is less than your configured maximum amount of failures for your proxy (i.e. [`max_fails`](/docs/json/apps/http/servers/routes/handle/reverse_proxy/health_checks/passive/max_fails/)), and that `num_requests` is less than or equal to your configured amount of maximum requests per upsteam (i.e. [`unhealthy_request_count`](/docs/json/apps/http/servers/routes/handle/reverse_proxy/health_checks/passive/unhealthy_request_count/) for the whole proxy, or [`max_requests`](/docs/json/apps/http/servers/routes/handle/reverse_proxy/upstreams/max_requests/) for individual upstreams).
```

Something like this, maybe?

mholt (changes requested):
How's this sound...

francislavoie:
Sounds good to me :+1: :grin:

mholt (approved):

PR #159, opened 2021-04-26
docs: Clarify `file_server` behaviour inside `handle_errors`

francislavoie:
This was part of https://github.com/caddyserver/website/pull/154 which was incorrect, but I realized I had this part in it which is still relevant to keep.


mholt (approved):
This is fine, but maybe this should be noted in the docs for `file_server` instead?

francislavoie:
Yeah, we could probably add it there too, but there was already an example here that used `file_server` inside of `handle_errors` so I think it makes sense here at least.

PR #160, opened 2021-04-28
docs: Add `php_fastcgi`'s `index off` option

francislavoie:
I somehow forgot to add this to the docs when I implemented it a while back. Someone asked a question on the forums where this was the solution, and I noticed it wasn't documented.


mholt (approved):
Nice catch, thanks!

PR #161, opened 2021-04-28
docs: Fix global timeouts labels

xpetit:
Make global timeouts labels correspond to what's written below (and what Caddy expects)


francislavoie:
Oops, you're right, thanks!

PR #167, opened 2021-05-28
Add a space between the icon and href in the card.

darkweak:
Add a space between the module icon and the module link is clearer imo.

e.g.
Capture d’écran 2021-05-29 à 01 05 11

Capture d’écran 2021-05-29 à 01 10 25


mholt:
Thanks; actually did this locally as part of [a much bigger change](https://github.com/caddyserver/website/commit/e132c5c45422c92fd8b9d385e3e2c01e25fa1857) though.

PR #168, opened 2021-05-29
docs: Revise documentaiton of trusted_* directives in tls/client_auth

favonia:
1. The current documentation is misleading. The `*_file` directives expects a regular PEM file, not (raw) DER or base64-encoded raw DER.
2. I further condensed the description in hope that it becomes more clear.

### DIFF
--- a/src/docs/markdown/caddyfile/directives/tls.md
+++ b/src/docs/markdown/caddyfile/directives/tls.md
@@ -86,12 +86,12 @@ tls [internal|<email>] | [<cert_file> <key_file>] {
 
 	Default: `require_and_verify` if any `trusted_ca_cert` or `trusted_leaf_cert` are provided; otherwise, `require`.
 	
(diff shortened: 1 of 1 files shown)


francislavoie:
Well, technically PEM _is_ base64-encoded DER, but also with line breaks at every 64 chars and ASCII armor around it. But specifying PEM is more accurate, so 👍

favonia:
> Well, technically PEM _is_ base64-encoded DER, but also with line breaks at every 64 chars and ASCII armor around it. But specifying PEM is more accurate, so +1

I actually deleted the armor and was confused until I read the code. :sweat_smile:

mholt (approved):
Okay, yeah, this is probably a good clarification. When embedding the certificate bytes directly into the config, they're base64-encoded DER _only_, not PEM (with armor and line breaks) because putting armor and line breaks into a config file is annoying.

When specifying a file name to load from, however, we expect a PEM format file because they are so common as files.

PR #172, opened 2021-06-14
docs: Add `propagation_timeout` to tls issuer options

francislavoie:
Added in https://github.com/caddyserver/caddy/pull/4178


mholt (approved):
Gracias!

PR #173, opened 2021-06-21
docs: Clarify `uri` syntax, replacement variables

francislavoie:
- The syntax definition was confusing because it was mixing one-of options and optional arguments which are only relevant for certain options.

- Add notes about replacement capture groups. Wording could probably be improved here. Point is, capture groups are supported, using Golang's syntax (doesn't use Caddy's placeholder syntax, which can be a gotcha)

- Add similar directives section, mainly to point out that `handle_path` can be used instead in many situations to save a line. The note about `rewrite` is kinda redundant with the small paragraph at the top though. Should we move that explanation down into `Similar directives`?

### DIFF
--- a/src/docs/markdown/caddyfile/directives/uri.md
+++ b/src/docs/markdown/caddyfile/directives/uri.md
@@ -11,10 +11,13 @@ This directive is distinct from [`rewrite`](rewrite) in that `uri` _partially_ c
 
 ## Syntax
 
+Multiple different operations are supported:
+
 ```caddy-d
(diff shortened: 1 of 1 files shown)


mholt (approved):
Thanks, this does seem clearer.

> Add similar directives section, mainly to point out that handle_path can be used instead in many situations to save a line. The note about rewrite is kinda redundant with the small paragraph at the top though. Should we move that explanation down into Similar directives?

I think it's fine tbh. :man_shrugging:  But you can if it's really bothering you, ha.

PR #175, opened 2021-06-23
docs: Add additional example to the `respond` directive

moritzdietz:
I always try to read all the examples on the docs to get a better understanding of what can be done and how it can be done.
Unbeknownst to me I could use backticks as delimiters to "escape" the escaping of characters which would need them otherwise.

Reference: https://twitter.com/mholt6/status/1407473895876136962

This PR adds an additional example which highlights this. I bet others will find this interesting and could benefit from it.

### DIFF
--- a/src/docs/markdown/caddyfile/directives/respond.md
+++ b/src/docs/markdown/caddyfile/directives/respond.md
@@ -49,3 +49,8 @@ respond /secret/* "Access denied" 403 {
 	close
 }
 ```
+
+To respond with a body containing double quotes, which would need escaping, you can make use of [backticks](/docs/caddyfile/concepts#tokens-and-quotes) to avoid escaping:
+```caddy-d
+respond `{"key": "value"}`
+```


francislavoie:
Thanks!

I think additionally, it would be helpful to link to https://caddyserver.com/docs/caddyfile/concepts#tokens-and-quotes which describes how that works.

moritzdietz:
Awesome! That makes it even more clearer; connecting the dots.
I'll add a section underneath pointing to that link and be right back.

francislavoie:
All you need to do really is link the word "backticks" to there. Don't really need an extra paragraph.

Also I'd clarify that by instead writing "if you need to respond with content including double quotes, then you may use backticks to avoid escaping them" or something to that effect

moritzdietz:
We just missed each other. 
>Also I'd clarify that by instead writing "if you need to respond with content including double quotes, then you may use backticks to avoid escaping them" or something to that effect

So basically writing the same text which can be found at the link you mentioned?
Which would be
```
Backtick strings are convenient when tokens contain quote literals, e.g. JSON text.
```

moritzdietz:
Alright I think this should be it now. Let me know if this works. I'll squash the commits then.

mholt:
I do want to be careful that examples don't turn into tutorials. This is probably fine, but does seem unrelated to what the directive is actually about. Do you need to set a Content-Type header?

moritzdietz:
I agree, it makes sense to limit the examples in general.
Adding examples for each and every use case is not ideal and brings a burden to the docs in case the behavior changes etc.

This PR was meant to be more like a suggestion. I also thought about adding a post to the caddy forum, showcasing this usage instead.
If you think it would fit there more, I'll take there.

> This is probably fine, but does seem unrelated to what the directive is actually about. Do you need to set a Content-Type header?

You mean the respond directive?
Why is it unrelated? What am I missing?

PR #176, opened 2021-06-29
Add instruction on how to run Caddy as a Windows service

TobiX:
This is a first stab at getting some simple instructions on how to run Caddy as a Windows service into the official doumentation.

Some things:
- Maybe put the XML file into https://github.com/caddyserver/dist somewhere?
- [WinSW](https://github.com/winsw/winsw/tree/master#usage) supports a stop command, but I noticed that `caddy stop` might not be universal if, for example, the admin port was changed. Leaving it out, caddy just receives a SIGINT from the service wrapper:
```json
{"level":"info","ts":1624970636.3611643,"msg":"shutting down","signal":"SIGINT"}
{"level":"warn","ts":1624970636.3611643,"msg":"exiting; byeee!! ðŸ‘‹","signal":"SIGINT"}
{"level":"info","ts":1624970636.3984616,"msg":"shutdown complete","signal":"SIGINT","exit_code":0}
```
- WinSW redirects STDERR & STDOUT into files, as a simple suggestion this is configured to roll daily. If caddy isn't writing that much to STDERR, it might be better to roll weekly or monthly?

### DIFF
--- a/src/docs/markdown/install.md
+++ b/src/docs/markdown/install.md
@@ -29,6 +29,7 @@ This page describes various methods for installing Caddy on your system.
 - [Ansible](#ansible)
 - [Scoop](#scoop)
 - [Termux](#termux)
+- [Windows service](#windows-service)
 
 
 ## Static binaries
@@ -222,4 +223,52 @@ _Note: This is a community-maintained installation method._
 <pre><code class="cmd">pkg install caddy</code></pre>
 
 [**View the Termux build.sh file**](https://github.com/termux/termux-packages/blob/master/packages/caddy/build.sh)
+
+
+## Windows service
+
+_Note: This is a community-maintained installation method._
+
+Install Caddy as a service on Windows with these instructions.
+
+**Requirements:**
+
+- `caddy.exe` binary that you [downloaded](/download) or [built from source](/docs/build)
+- Any exe from the latest release of the
+  [WinSW](https://github.com/winsw/winsw/releases/latest) service wrapper (Stay
+  on a v2.x release for now)
+
+Put all files into a service directory. In the following examples, we use `c:\caddy`.
+
+Rename the WinSW exe file to `caddy-service.exe`.
+
+Add a `caddy-service.xml` in the directory:
+
+<pre><code class="cmd">&lt;service>
+  &lt;id>caddy&lt;/id>
+  &lt;!-- Display name of the service -->
+  &lt;name>Caddy Web Server (powered by WinSW)&lt;/name>
+  &lt;!-- Service description -->
+  &lt;description>Caddy Web Server (https://caddyserver.com/)&lt;/description>
+  &lt;executable>%BASE%\caddy.exe&lt;/executable>
+  &lt;arguments>run&lt;/arguments>
+  &lt;log mode="roll-by-time">
+    &lt;pattern>yyyy-MM-dd&lt;/pattern>
+  &lt;/log>
+&lt;/service>
+</code></pre>
+
+You can now install the service using:
+<pre><code class="cmd">caddy-service install</code></pre>
+
+You might want to start the Windows Services Console to see if the service is runnnig correctly:
+<pre><code class="cmd">services.msc</code></pre>
+
+Be aware that Windows services cannot be reloaded, so you have to tell caddy directly to relaod:
+<pre><code class="cmd">caddy reload</code></pre>
+
+Restarting is possible via the normal Windows services commands.
+
+For customizing the service wrapper, see the [WinSW documentation](https://github.com/winsw/winsw/tree/master#usage)
+
 


src/docs/markdown/install.md
```diff
@@ -211,3 +212,50 @@ _Note: This is a community-maintained installation method._
 <pre><code class="cmd">scoop install caddy</code></pre>
 
 [**View the Scoop manifest**](https://github.com/ScoopInstaller/Main/blob/master/bucket/caddy.json)
+
+
+## Windows service
```
mholt, line 228:
```suggestion
## WinSW service
```

Since this uses WinSW instead of native Windows, right?

mholt (changes requested):
Thanks for this; I guess it'll be handy to have some instructions for Windows. This appears to rely on a third party tool, WinSW, which is fine as this is documented as a community-contributed installation method.

At some point I think it would be good to have truly "official" instructions for installing as a Windows service that doesn't require third party tools (doesn't Windows have a Task Scheduler, for example? I don't know if that's the right way to make a service).

> Maybe put the XML file into https://github.com/caddyserver/dist somewhere?

Probably only if it's an official distribution method, but since it relies on (looks like arbitrary) third party software, probably not.

> WinSW supports a stop command, but I noticed that caddy stop might not be universal if, for example, the admin port was changed.

The stop command can be given an `--address` flag if the port is different from the default.

francislavoie:
This looks alright. I typically use [NSSM](https://nssm.cc/) or recommend using that when the question is asked on https://caddy.community. Setup looks like this:

```
.\nssm.exe install caddy C:\caddy\caddy.exe
.\nssm.exe set caddy AppParameters "run --config Caddyfile --watch"
.\nssm.exe set caddy AppDirectory C:\caddy
.\nssm.exe set caddy AppExit Default Restart
.\nssm.exe set caddy AppNoConsole 1
.\nssm.exe set caddy AppStdout C:\caddy\stdout.log
.\nssm.exe set caddy AppStderr C:\caddy\stderr.log
.\nssm.exe set caddy AppTimestampLog 1
.\nssm.exe set caddy Description "Caddy Server"
.\nssm.exe set caddy DisplayName Caddy
.\nssm.exe set caddy ObjectName LocalSystem
.\nssm.exe set caddy Start SERVICE_AUTO_START
.\nssm.exe set caddy Type SERVICE_INTERACTIVE_PROCESS
```

I suppose using an XML definition for the service is a bit more portable than a list of commands.

src/docs/markdown/install.md
francislavoie: WinSW is a tool to spin up a _Windows service_. Services are a concept innate to Windows.

francislavoie: The alternative (to not require some third party tool) is to write and give a powershell script to do the job, but that's very not fun to maintain or recommend to use. I rather recommend some third party tool which gives it a nicer API (like NSSM or WinSW; TIL of WinSW)

mholt: How do you do it without a third party tool? (Not that I'm against documenting WinSW, but for official install instructions we should use native facilities.)

mholt:
Huh, the NSSM one does look a lot simpler though.

francislavoie:
It's basically the same amount of work tbh. NSSM can be sorta funky cause it has this janky GUI that looks pretty dated. WinSW looks like a more modern CLI API, at a glance. NSSM isn't my favourite thing.

src/docs/markdown/install.md
TobiX: Software running as a Windows service must adhere to some interfaces, WinSW or nssm are basically wrappers to implement these features and map them to more "standard"/UNIXy process control. WinSW was born in the Jenkins project (IIRC), because implementing those service interfaces in Java isn't that easy (maybe even impossible? I think some features need to be inside the exe file that is registered as the service?)

TobiX:
I'm a little bit opposed to nssm, since it is often mis-detected as malware by some AV. (Also, the latest release is from 2014 :grinning:)

src/docs/markdown/install.md
TobiX: Maybe just support the service interface directly? I just found https://github.com/kardianos/minwinsvc which promises Windows service support with no code changes :+1:

TobiX:
Some more thoughts:

To avoid those external tools: Would you accept a patch implementing the service interface in Caddy?

Looking at the examples in https://pkg.go.dev/golang.org/x/sys/windows/svc, this shouldn't be that complicated (well, I haven't written that much Go code yet, so this would be a learning experience for me). The minimal implementation would just need to:
- Shortly after starting, signal the service manager we are running
- When the service manager signals us to stop, stop
(Incidently, that is everything that https://github.com/kardianos/minwinsvc does, so maybe including that in the right place might be the way to go if you don't want all this Windows-specific code inside Caddy?)

More fancy integration would involve (not needed for the first integration):
- Providing a command line option to install/uninstall the service. This would be useful since doing this manually is a bit involved and error-prone, for example (taken from the [SVN book](http://svnbook.red-bean.com/en/1.8/svn.serverconfig.svnserve.html#svn.serverconfig.svnserve.invoking.winservice), yes all those spaces are intentional):
```cmd
sc create svn
        binpath= "C:\svn\bin\svnserve.exe --service -r C:\repos"
        displayname= "Subversion Server"
        depend= Tcpip
        start= auto
```
- Logging to the Windows event log instead of STDOUT/STDERR (those are just discarded for Windows services)


PS: systemd has similar interfaces (https://manpages.debian.org/buster/libelogind-dev-doc/sd_notify.3.en.html) for communication of the service with the service manager, so maybe a generic interface would be useful...

mholt:
Am now thinking about combining this with #181. What would you think of that, a new "System service" docs page?

> To avoid those external tools: Would you accept a patch implementing the service interface in Caddy?

Hmm, possibly, but would need to see a PoC to be sure.

> PS: systemd has similar interfaces (https://manpages.debian.org/buster/libelogind-dev-doc/sd_notify.3.en.html) for communication of the service with the service manager, so maybe a generic interface would be useful...

I believe we implement those: https://github.com/caddyserver/caddy/tree/master/notify -- bonus points if that interface can be re-utilized for Windows, or maybe make them the same for each platform so we don't have to worry about that from Caddy core.

mholt:
@TobiX Want to resolve the merge conflict and then we can merge this in so you get credit, and then #181 can update to move it into a new page?

mholt (approved):
Thanks for the contribution! We'll merge this in so you get credit, then combine with our new services docs page.

PR #177, opened 2021-06-29
Remove usage of apt-key

TobiX:
This replaces the usage of `apt-key` by putting the ASCII armored key into `/etc/apt/trusted.gpg.d`, see https://blog.jak-linux.org/2021/06/20/migrating-away-apt-key/ for the deprecation notice for `apt-key`.

ASCII armored keys are supported since apt 1.4, so this will not work on the (still supported) Ubuntu versions 14.04 (trusty) and 16.04 (xenial). On the other hand, those releases don't use systemd, so most of the other instructions won't work either.


francislavoie (approved):
Seems fine to me 👍

mholt (approved):
LGTM, thanks, had no idea.

PR #181, opened 2021-07-04
docs: Add new "Keep Caddy Running" page, moved from install.md

francislavoie:
I found that the way we were documenting how to use Caddy as a systemd service wasn't adequate - there are many common tasks we weren't properly covering in `install.md`, and it also didn't feel like the appropriate place to expand on it, since usage is a separate concern from installation.

So I decided to make a new "Linux Service" page for this, where we can elaborate some more.

I'm not sure I covered everything necessary here, but I think this should be a good start.

Also tagging @carlwgeorge because you've helped me understand a lot of these details over the past year :smile:

### DIFF
--- a/src/docs/markdown/install.md
+++ b/src/docs/markdown/install.md
@@ -14,7 +14,6 @@ This page describes various methods for installing Caddy on your system.
 - [Arch Linux, Manjaro, Parabola](#arch-linux-manjaro-parabola)
 - [Docker](#docker)
 - [DigitalOcean](#digitalocean)
-- [Linux service](#linux-service)
 
 <aside class="tip">
     Our <a href="https://github.com/caddyserver/dist">official packages</a> come only with the standard modules. If you need third-party plugins, <a href="/docs/build#xcaddy">build from source with <code>xcaddy</code></a> or use <a href="/download">our download page</a>.
@@ -34,7 +33,7 @@ This page describes various methods for installing Caddy on your system.
 
 ## Static binaries
 
-Simply downloading a Caddy binary does not <a href="#linux-service">install it as a service</a>, but can be useful in dev or when upgrading an existing installation.
+Simply downloading a Caddy binary does not [install it as a service](/docs/service#manual-installation), but can be useful in dev or when upgrading an existing installation.
 
 - [**View releases on GitHub**](https://github.com/caddyserver/caddy/releases) (expand "Assets")
 - [**Use our download page**](/download)
@@ -42,9 +41,7 @@ Simply downloading a Caddy binary does not <a href="#linux-service">install it a
 
 ## Debian, Ubuntu, Raspbian
 
-Installing this package automatically starts and runs Caddy for you as a systemd service named `caddy` using our official [`caddy.service`](https://github.com/caddyserver/dist/blob/master/init/caddy.service) unit file.
-
-This package also comes with a `caddy-api` systemd service using our official [`caddy-api.service`](https://github.com/caddyserver/dist/blob/master/init/caddy-api.service) unit file, which is disabled by default. If you plan to configure Caddy solely through its [API](/docs/api), then you should disable the `caddy` service, and enable the `caddy-api` service.
+Installing this package automatically starts and runs Caddy as a [systemd service](/docs/service) named `caddy`, and also comes with a `caddy-api` service which is _not_ enabled by default, should you need it.
 
 Stable releases:
 
@@ -64,9 +61,12 @@ Testing releases (includes betas and release candidates):
 
 [**View the Cloudsmith repos**](https://cloudsmith.io/~caddy/repos/)
 
-If you wish to use the packaged support files (systemd services, bash completion and default configuration) with a custom Caddy build, instructions can be found [here](https://caddyserver.com/docs/build#package-support-files-for-custom-builds-for-debian-ubuntu-raspbian).
+If you wish to use the packaged support files (systemd services, bash completion and default configuration) with a custom Caddy build, instructions can be [found here](https://caddyserver.com/docs/build#package-support-files-for-custom-builds-for-debian-ubuntu-raspbian).
+
 
 ## Fedora, RedHat, CentOS
+
+This package comes with both of Caddy's [systemd service](/docs/service) unit files, but does not enable them by default.
 
 Fedora or RHEL/CentOS 8:
 
@@ -82,7 +82,10 @@ RHEL/CentOS 7:
 
 [**View the Caddy COPR**](https://copr.fedorainfracloud.org/coprs/g/caddy/caddy/)
 
+
 ## Arch Linux, Manjaro, Parabola
+
+This package comes with both of Caddy's [systemd service](/docs/service) unit files, but does not enable them by default.
 
 <pre><code class="cmd"><span class="bash">pacman -Syu caddy</span></code></pre>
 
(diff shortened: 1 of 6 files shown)


francislavoie:
Also tagging @shibumi since `pacman` also ships with the service files.

mholt:
Thinking ahead here, what if we had a "System service" page that included Linux, and in the future, Windows?

francislavoie:
Yeah, we could.

mholt:
What if we did that, then -- we could combine #176 and this into the same document?

francislavoie:
Sure. /cc @TobiX

Do we want to merge that PR first so his authorship is retained, then I can move it in this PR after? Probably the simplest thing to do I figure.

mholt:
I'm OK with that, yeah.

src/docs/markdown/install.md
```diff
@@ -63,10 +60,13 @@ Testing releases (includes betas and release candidates):
 
 [**View the Cloudsmith repos**](https://cloudsmith.io/~caddy/repos/)
 
-If you wish to use the packaged support files (systemd services, bash completion and default configuration) with a custom Caddy build, instructions can be found [here](https://caddyserver.com/docs/build#package-support-files-for-custom-builds-for-debian-ubuntu-raspbian).
+If you wish to use the packaged support files (systemd services, bash completion and default configuration) with a custom Caddy build, instructions can be [found here](https://caddyserver.com/docs/build#package-support-files-for-custom-builds-for-debian-ubuntu-raspbian).
```
carlwgeorge:
```suggestion
If you wish to use the packaged support files (systemd services, bash completion and default configuration) with a custom Caddy build, instructions can be [found here](https://caddyserver.com/docs/build#package-support-files-for-custom-builds-for-debianubunturaspbian).
```

src/docs/markdown/service.md
```diff
@@ -0,0 +1,121 @@
...
+
+We provide two different systemd unit files that you can choose between, depending on your usecase:
+
+- [**`caddy.service`**](https://github.com/caddyserver/dist/blob/master/init/caddy.service) if you configure Caddy with a [Caddyfile](/docs/caddyfile). If you prefer to use a JSON config file, you may [override](#overrides) the `ExecStart` and `ExecReload` commands.
+
+- [**`caddy-api.service`**](https://github.com/caddyserver/dist/blob/master/init/caddy-api.service) if you configure Caddy solely through its [API](/docs/api). This service uses the [`--resume`](/docs/command-line#caddy-run) option which will start Caddy using the `autosave.json` which is [persisted](/docs/json/admin/config/) by default.
+
+They are very similar, but differ in the `ExecStart` and `ExecReload` commands to accommodate the workflows.
+
+If you need to switch between the services, you should disable the previous one before enabling the other. For example, to switch from the `caddy` service to the `caddy-api` service:
```
carlwgeorge:
```suggestion
If you need to switch between the services, you should disable and stop the previous one before enabling and starting the other. For example, to switch from the `caddy` service to the `caddy-api` service:
```

src/docs/markdown/service.md
```diff
@@ -0,0 +1,121 @@
...
+<pre><code class="cmd"><span class="bash">sudo systemctl disable caddy</span>
+<span class="bash">sudo systemctl enable caddy-api</span>
+<span class="bash">sudo systemctl start caddy-api</span></code></pre>
```
carlwgeorge:
```suggestion
<pre><code class="cmd"><span class="bash">sudo systemctl disable --now caddy</span>
<span class="bash">sudo systemctl enable --now caddy-api</span></code></pre>
```

src/docs/markdown/service.md
```diff
@@ -0,0 +1,121 @@
...
+
+**Double-check the `ExecStart` and `ExecReload` directives.** Make sure the binary's location and command line arguments are correct for your installation! For example: if using a config file, change your `--config` path if it is different from the defaults.
+
+The usual place to save the service file is: `/etc/systemd/system/caddy.service`
+
+After saving your service file, you can start the service for the first time with the usual systemctl dance:
+
+<pre><code class="cmd"><span class="bash">sudo systemctl daemon-reload</span>
+<span class="bash">sudo systemctl enable caddy</span>
+<span class="bash">sudo systemctl start caddy</span></code></pre>
```
carlwgeorge:
```suggestion
<span class="bash">sudo systemctl enable --now caddy</span></code></pre>
```

src/docs/markdown/service.md
```diff
@@ -0,0 +1,121 @@
...
+```systemd
+[Service]
+Environment="CF_API_TOKEN=super-secret-cloudflare-tokenvalue"
+```
+
+Or, for example if you need to change the config file from the default of the Caddyfile, to instead using a JSON file:
+```systemd
+[Service]
+ExecStart=/usr/bin/caddy run --environ --config /etc/caddy/caddy.json
+ExecReload=/usr/bin/caddy reload --config /etc/caddy/caddy.json
```
carlwgeorge: `Exec*` directives are allowed to be specified multiple times.  An override with just another `Exec*` line will cause both the original and the new one to be run.  They has to be "reset" with empty directives to properly override them.

https://www.freedesktop.org/software/systemd/man/systemd.service.html#ExecStart=

```suggestion
ExecStart=
ExecStart=/usr/bin/caddy run --environ --config /etc/caddy/caddy.json
ExecReload=
ExecReload=/usr/bin/caddy reload --config /etc/caddy/caddy.json
```

carlwgeorge (changes requested):
Looks mostly good, just a few small corrections needed.

src/docs/markdown/service.md
francislavoie: TIL! Thanks

src/docs/markdown/service.md
francislavoie: Oh that's nice, thanks!

src/docs/markdown/service.md
```diff
@@ -0,0 +1,121 @@
+---
+title: Linux Service
```
mholt:
```suggestion
title: System Service
```

And maybe rename the file to `system-service.md`? I dunno. Maybe `service.md` is fine.

src/docs/markdown/service.md
```diff
@@ -0,0 +1,121 @@
...
+---
+
+Linux Service
```
mholt: Make a new h1, "System service", and then make this a h2, and all the h2's into h3's, etc.

src/includes/docs-nav.html
```diff
@@ -46,6 +46,7 @@
 		<li><a href="/docs/logging">How Logging Works</a></li>
 		<li><a href="/docs/metrics">Monitoring Caddy</a></li>
 		<li><a href="/docs/architecture">Caddy Architecture</a></li>
+		<li><a href="/docs/service">Linux Service</a></li>
```
mholt:
```suggestion
		<li><a href="/docs/service">System service</a></li>
```

And I wonder if this should be in the "Get Caddy" section, but I can see why it's good in "Articles" too as it's not really about "getting" Caddy...

mholt (changes requested):
Thanks for working on this; making progress!

src/docs/markdown/service.md
francislavoie: Yeah, I'll do another pass on this when pulling in the Windows stuff.

I think `service.md` is fine for brevity of the URL 😅

src/includes/docs-nav.html
francislavoie: All the other "articles" use Title Case, should we not do the same for this page? i.e. `System Service`

mholt: Yes, I suppose we should. :+1:

mholt:
@francislavoie #176 was merged, so we can add those instructions to the new docs page.

francislavoie:
Great, I'll work on that this week

src/docs/markdown/running.md
```diff
@@ -0,0 +1,214 @@
+---
```
mholt, line 1: What if we call this file `run.md` instead? (`/docs/run`)

mholt (approved):
LGTM, just a question about a file name. Thanks!

src/docs/markdown/running.md
francislavoie: Given the title is "Keep Caddy _Running_", feels better to have an actual word from the title as the URL? 🤔

mholt:
Thanks Francis, this is good -- though if I find a better term, I might switch it to use that. Appreciate it!

geoidesic:
So... where do we find this wonderful page?

mholt:
@geoidesic "Keep Caddy Running" in our docs navigation. https://caddyserver.com/docs/running

geoidesic:
@mholt. Tx. From this thread: (https://caddy.community/t/can-i-set-the-user-manually-instead-of-caddy-for-compiled-deb-files/13191/2), I thought it was going to talk about setting the caddy linux user. I'm struggling to get the folder permissions on my server correct for CakePHP.

Basically the `webroot/tmp/cache` folder needs to be writeable by the web app, but it's not. I have created a user. I currently have this:

```
[root@app-staging tmp]# pwd
/var/www/html/app/tmp
[root@app-staging tmp]# ls -la
drwxrwxrwx  5 app-staging caddy    51 Jul  2 17:31 cache
[root@app-staging tmp]# ls -la cache/
drwxr-xrwx 2 app-staging caddy  6 Jul  2 17:36 models
```
But I'm getting errors: 
```
Warning (512): /var/www/html/app/tmp/cache/models/ is not writable [CORE/src/Cache/Engine/FileEngine.php, line 427]
```

PR #185, opened 2021-08-09
Use correct response header in CORS example

miso-belica:
`Access-Control-Request-Method` is a request header
sent by the client (browser, ...) but we need to send
response header with all allowed methods.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Request-Method
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Methods

----------------------------

Also, I was not able to include `import cors` multiple times like this but I found out a solution. I can add it in the new commit if needed.

```
(cors) {
	@origin{args.0} header Origin {args.0}
	header @origin{args.0} Access-Control-Allow-Origin "{args.0}"
	header @origin{args.0} Access-Control-Request-Method GET
	header @origin{args.0} Access-Control-Allow-Methods "OPTIONS,HEAD,GET,POST,PUT,PATCH,DELETE"
}
example.com {
	import cors example.com
	import cors another.example.com
}
```

### DIFF
--- a/src/docs/markdown/caddyfile/directives/import.md
+++ b/src/docs/markdown/caddyfile/directives/import.md
@@ -32,7 +32,7 @@ Import a snippet that sets CORS headers using an import argument:
 (cors) {
 	@origin header Origin {args.0}
 	header @origin Access-Control-Allow-Origin "{args.0}"
-	header @origin Access-Control-Request-Method GET
+	header @origin Access-Control-Allow-Methods "OPTIONS,HEAD,GET,POST,PUT,PATCH,DELETE"
 }
 
 example.com {


mholt (approved):
Ah, thanks for the fix. But why do we need to add all the methods? Isn't it best practice to only list the allowed methods, and make them as narrow as possible? I don't know of any endpoints that use ALL of the HTTP methods...

miso-belica:
Hi, headers are sent for the whole application, not just one endpoint and usually you use all methods in the application. Also, it's just an example in the documentation that can be adjusted if someone is not using PATCH or other methods for example.

Another possibility is to dynamically send just one method but it could be tricky with preflight requests. This is the easiest way :)

francislavoie:
Why not just set it to `Access-Control-Allow-Methods: *`? Bit simpler to read, effectively the same thing, less bytes on the response.

Edit: Wow, apparently Safari doesn't support `*`. Safari is _the worst_. https://caniuse.com/mdn-http_headers_access-control-allow-methods_wildcard

miso-belica:
Hi, I didn't know Safari doesn't support it but I agree it's the worst 😀 But the reason why I enumerated all methods instead of using * is that * does not support requests with credentials (cookies for example). It's in the documentation I sent in the PR description. Basically the same situation as with `Access-Control-Allow-Origin: *`. I know this is not optimal for every use-case but it's the simplest implementation that works in all cases.

PR #187, opened 2021-08-12
js: Fix "Cannot read property 'split' of null"

### LINKED ISSUE #186: caddyserver.com/download broken when user agent not recognised
Perhaps I'm missing something, because this feels like an issue which would have been caught/reported instantly, but the page at https://caddyserver.com/download appears to have no functionality at all other than downloading a single linux/amd64 binary despite many more options being seemingly available. I've found the following issues:

- Platform selection box does nothing. The button always downloads `caddy_linux_amd64`.
- Modules/plugins cannot be selected at all.
- The search box does nothing.

These make the entire page effectively a download button, because all other functionality is unavailable.

I'm on `openSUSE Tumbleweed` with `Brave Browser 1.28.105 (Chromium: 92.0.4515.131)`.

francislavoie:
Fix #186 

In certain cases where the platform cannot be automatically determined, we should at least populate a default so that the download page doesn't break.

This uses https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment to save a few lines for the split.

### DIFF
--- a/src/download.html
+++ b/src/download.html
@@ -23,8 +23,7 @@
 				<div>
 					<div>
 						<b>Platform:</b>
-						<select size="1" id="platform">
-							<option></option>
+						<select size="1" id="platform" selected="linux-amd64">
 							<option value="dragonfly-amd64">Dragonfly amd64</option>
 							<option value="freebsd-amd64">FreeBSD amd64</option>
 							<option value="freebsd-arm-6">FreeBSD arm 6</option>
--- a/src/resources/js/download.js
+++ b/src/resources/js/download.js
@@ -169,11 +169,11 @@ $(function() {
 	});
 })
 
-// autoPlatform choooses the platform in the list that best
+// autoPlatform chooses the platform in the list that best
 // matches the user's browser, if it's available.
 function autoPlatform() {
 	// assume 32-bit linux, then change OS and architecture if justified
-	var os = "linux", arch = "386", arm = "";
+	var os = "linux", arch = "amd64", arm = "";
 
 	// change os
 	if (/Macintosh/i.test(navigator.userAgent)) {
@@ -215,11 +215,15 @@ function autoPlatform() {
 }
 
 function getDownloadLink() {
+	// make sure we at least have a default,
+	// in the case that autoPlatform() failed
+	var platformString = $('#platform').val();
+	if (!platformString) {
+		platformString = "linux-amd64"
+	}
+
 	// get platform components
-	var platformParts = $('#platform').val().split("-");
-	var os = platformParts[0];
-	var arch = platformParts[1];
-	var arm = platformParts.length > 2 ? platformParts[2] : "";
+	var [os, arch, arm = ""] = platformString.split("-");
 
 	var qs = new URLSearchParams();
 	if (os)   qs.set("os", os);


0x1a8510f2:
Some related things I found while looking into this myself which may be worth including in this PR:

- [`autoPlatform()` initially assumes `linux-386`](https://github.com/caddyserver/website/pull/187/files#diff-f79367e3b2cbd2b261fe6e9967fc9f97dcd917d1683ce6dae35a02768db1a06cR175-R176) but that's not an option in the select box and it doesn't match the default in this PR.
- The [empty option](https://github.com/caddyserver/website/blob/master/src/download.html#L27) should probably be removed from the `<select>` in `download.html` since the JS already defines a default (see above) and an option must be selected before downloading anyway.
- A `selected` could be added to the default option in the HTML.
- Typo at https://github.com/caddyserver/website/blob/master/src/resources/js/download.js#L172 :P

francislavoie:
Good points. I've made those adjustments. Thanks @0x1a8510f2!

src/resources/js/download.js
```diff
@@ -169,11 +169,11 @@ $(function() {
 	});
 })
 
-// autoPlatform choooses the platform in the list that best
+// autoPlatform chooses the platform in the list that best
 // matches the user's browser, if it's available.
 function autoPlatform() {
 	// assume 32-bit linux, then change OS and architecture if justified
-	var os = "linux", arch = "386", arm = "";
+	var os = "linux", arch = "amd64", arm = "";
```
mholt, line 176: If we change this default then we need a way to detect 32-bit and set it below.

src/download.html
```diff
@@ -23,8 +23,7 @@
 				<div>
 					<div>
 						<b>Platform:</b>
-						<select size="1" id="platform">
-							<option></option>
+						<select size="1" id="platform" selected="linux-amd64">
```
mholt, line 26: If we don't know the platform, I don't think it's a good idea to select one by default.

mholt (changes requested):
Thanks, the destructured assignment is a nice improvement. I think there are some subtlties here we need to address though.

src/download.html
francislavoie: I'm pretty sure `linux-amd64` is by far the most used platform for Caddy (probably 65%+? https://github.com/caddyserver/caddy/releases assets suggests as much) so I don't think there's any harm in defaulting to that if we can't determine the platform otherwise. The user can pick a different one if they prefer. Smooths out the UX to at least have a sensible default.

mholt: I can see it now: people complaining that they can't run the binary they downloaded because it's the wrong platform. What if we leave it empty and show an error message if a platform is not selected?

(FWIW, I don't care too strongly about this. Just want to be thorough.)

src/resources/js/download.js
francislavoie: I don't think we need to worry about precision of the detection. It's pretty much best-guess. The User-Agent header is pretty unreliable and some browsers are starting to intentionally give bad values for privacy reasons.

That said I realize that the comment above this line is wrong now, I'll fix that.

mholt: How do we auto-select 32-bit platforms with this change?

0x1a8510f2: From what I remember there isn't even a 32-bit option in the HTML select box

src/download.html
0x1a8510f2: I feel like people who set up webservers can be trusted to understand an exec format error. Then again, one should never overestimate the intelligence of users :P still, personally I think it's fine as-is. A blank option would need a lot more handling/checking which is why I suggested removing it in the first place.

src/resources/js/download.js
mholt: Ah, that's true. I removed them with the new site since 386 was our least popular platform.

mholt (approved):
I suppose we can give this a shot and then make changes later. Thanks for the fix!

PR #189, opened 2021-08-26
docs: Updates for Caddy v2.4.4

francislavoie:


src/docs/markdown/caddyfile/directives/reverse_proxy.md
```diff
@@ -174,8 +174,6 @@ By default, Caddy passes thru incoming headers to the backend&mdash;including th
 - It adds or augments the [X-Forwarded-For](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For) header field.
 - It sets the [X-Forwarded-Proto](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto) header field.
 
-Since these header fields are only de-facto standards, Caddy may stop setting them implicitly in the future if the standardized [Forwarded](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded) header field becomes more widely adopted.
```
francislavoie, line 177: I'm removing this, because frankly, it's confusing. The docs shouldn't speculate about the future, it should just talk about what Caddy does _now_. I think it's very unlikely we'll ever remove `X-Forwarded-For`, because pretty much _everything_ uses those. There really doesn't seem to be any industry appetite for moving to `Forwarded` (it's a more complicated format to parse, kinda does too much).

src/docs/markdown/caddyfile/directives/reverse_proxy.md
```diff
@@ -309,31 +309,31 @@ reverse_proxy /api/* node1:80 node2:80 node3:80 {
...
 
 ```caddy-d
 reverse_proxy https://example.com {
-	transport http {
-		dial_timeout 2s
-		tls_timeout  2s
-	}
+	header_up Host {upstream_hostport}
+	header_up X-Forwarded-Host {host}
 }
```
francislavoie: I reorganized these to try to help clarify the actually usecase where `header_up Host {upstream_hostport}` is most commonly used. This is probably a bit too wordy though, I think I should move some of this copy to the `Headers` section above. 🤔

src/docs/markdown/logging.md
```diff
@@ -76,7 +76,8 @@ Now compare an equivalent structured log message from Caddy, encoded as JSON and
 			"server_name": "example.com"
 		}
 	},
-	"latency": 0.000014711,
+	"user_id": "",
+	"duration": 0.000014711,
```
francislavoie, line 80: The logs haven't used `"latency"` for a very long time now, it's been `"duration"`. Was a bit misleading.

src/download.html
```diff
@@ -50,7 +50,7 @@
 							<option value="windows-amd64">Windows amd64</option>
 							<option value="windows-arm-6">Windows arm 6</option>
 							<option value="windows-arm-7">Windows arm 7</option>
-							<!-- <option value="windows-arm64">Windows arm64</option> TODO: Go 1.17 - https://github.com/golang/go/issues/36439 -->
+							<option value="windows-arm64">Windows arm64</option>
```
francislavoie, line 53: I think this is ready to go along with v2.4.4. I haven't tested the cross-compile with `xcaddy` yet, if you want to try that on the build server as a sanity check, that'd be great

src/docs/markdown/caddyfile/directives/file_server.md
```diff
@@ -30,6 +31,7 @@ file_server [<matcher>] [browse] {
...
 - **precompressed** is the list of encoding formats to search for precompressed sidecar files. Arguments are an ordered list of encoding formats to search for precompressed sidecar files. Supported formats are `gzip`, `zstd` and `br`.
 - **status** is an optional status code override to be used when writing the response. Particularly useful when responding to a request with a custom error page. Can be a 3-digit status code, For example: `404`. Placeholders are supported. By default, the written status code will typically be `200`, or `206` for partial content.
+- **disable_canonical_uris** disables the default behaviour of redirecting to add a trailing slash if the request path is a directory, or remove the trailing slash if the request path is a file. Note that by default, canonicalization does not happen if the filename was underwent an internal rewrite (possibly via the `rewrite` directive), to avoid clobbering an explicit rewrite with implicit behaviour.
```
mholt: Canonicalization doesn't happen if the _end_ of a filename underwent an internal rewrite, right?

src/docs/markdown/caddyfile/directives/reverse_proxy.md
mholt: Have multiple users been confused by this?

I don't feel strongly either way, just wondering.

src/docs/markdown/command-line.md
```diff
@@ -67,7 +67,13 @@ The ellipses `...` indicates a continuation, i.e. one or more parameters.
 
 - **[caddy upgrade](#caddy-upgrade)**
   Upgrades Caddy to the latest release
-  
+
+- **[caddy add-package](#caddy-add-package)**
+  Upgrades Caddy to the latest release, with additional plugins added
```
mholt, line 72: Hmm, didn't we make it so that add-package and remove-package don't change/upgrade the Caddy version?

src/docs/markdown/caddyfile/directives/reverse_proxy.md
```diff
@@ -155,7 +155,7 @@ Passive health checks happen inline with actual proxied requests:
 
 The proxy **buffers responses** by default for wire efficiency:
 
-- **flush_interval** is a [duration value](/docs/conventions#durations) that defines how often Caddy should flush the buffered response body to the client. Set to -1 to disable buffering. It is set to -1 automatically for requests that have a `text/event-stream` response or for HTTP/2 requests where the Content-Length is unspecified.
+- **flush_interval** is a [duration value](/docs/conventions#durations) that adjusts how often Caddy should flush the response buffer to the client. By default, no periodic flushing is done. A negative value disables response buffering, and flushes immediately after each write to the client. This option is ignored when the upstream's response is recognized as a streaming response, or if its content length is `-1`; for such responses, writes are flushed to the client immediately.
```
mholt, line 160: "no periodic flushing is done" and "writes are flushed to the client immediately" are different or the same? (We should try to be clear about this in the docs.) In addition, I might also suggest putting the default part at the end.
```

**Review by mholt — CHANGES_REQUESTED** [review]
Thanks for beating me to it. Mostly LGTM. I agree the part about the `header_up Host` change is still a bit confusing though... hmm.

File: src/docs/markdown/caddyfile/directives/file_server.md (hunk 5 above)
Comment (francislavoie) [reply src/docs/markdown/caddyfile/directives/file_server.md]: The wording is taken pretty much verbatim from the godoc, so :man_shrugging: I didn't review the implementation

Comment (francislavoie) [reply src/docs/markdown/caddyfile/directives/file_server.md]: You wrote it here:

https://github.com/caddyserver/caddy/commit/191dc86f9e40c12517a133cf2181631c9adec791

File: src/docs/markdown/caddyfile/directives/reverse_proxy.md (hunk 1 above)
Comment (francislavoie) [reply src/docs/markdown/caddyfile/directives/reverse_proxy.md:177]: Yeah there were some users that asked "I shouldn't rely on this because it might go away later". I didn't dig on the forum to find an example, I just read it again while going through and found it misleading. It's not really valid to mention this because it's not actually deprecated or whatever, and we don't actually have any plans to change this.

File: src/docs/markdown/command-line.md (hunk 6 above)
Comment (francislavoie) [reply src/docs/markdown/command-line.md:72]: I don't think so? Didn't read to me like that was the case in the code. Could you check?

File: src/docs/markdown/caddyfile/directives/reverse_proxy.md (hunk 7 above)
Comment (francislavoie) [reply src/docs/markdown/caddyfile/directives/reverse_proxy.md:160]: I essentially copied the wording from stdlib when I did that syncing PR :thinking: 

I kept the wording as close as possible to the godoc here.

```go
	// Adjusts how often to flush the response buffer. By default,
	// no periodic flushing is done. A negative value disables
	// response buffering, and flushes immediately after each
	// write to the client. This option is ignored when the upstream's
	// response is recognized as a streaming response, or if its
	// content length is -1; for such responses, writes are flushed
	// to the client immediately.
	FlushInterval caddy.Duration `json:"flush_interval,omitempty"`
```

src/docs/markdown/caddyfile/directives/file_server.md
mholt: Right, but crucially, canonicalization can still happen if there's a rewrite: just not on the end of the path.

francislavoie: > Note that by default, canonicalization does not happen if the filename was underwent an internal rewrite

It does say as much. How would you write it otherwise (aside from the grammatical error that I just noticed lmao)?

francislavoie: I adjusted the wording.

src/docs/markdown/command-line.md
mholt: Hmm I guess you're right. Before too long, I'd like to make it so that add-package and remove-package don't change the current Caddy version.

mholt (approved):
Thank you for the diligence, Francis! Makes things much easier.

PR #190, opened 2021-08-26
docs: Dynamically turn subdirectives into links to anchors

francislavoie:
This makes navigating some of the bigger Caddyfile pages much nicer. This is a somewhat more generalized solution to adding links to subdirectives throughout the pages.

I only turned this on for the longer pages, because it's less beneficial for shorter pages since the subdirectives fit on the same screen at code blocks.

Essentially how it works is I added `<span id="subdirective"/>` anchors beside each bullet item for subdirectives to make an invisible anchor point that we can jump to. Then, on page load, I use `$('article *[id]')` to get all of the IDs on elements within the article body (includes any `h2` headers and such), which acts as an allow-list. Then I grab all the `$('pre.chroma .k')` elements which are the directive or subdirectives as per our syntax highlighting. Then it linkifies them if they have an ID on the page that matches.

I moved the script for the `options` page to a script block in the `.md` itself since it was kinda weird to have page-specific code in `docs.js`, we'd need to check the current request path which wasn't ideal.

I also added a bit of edgecase magic for other pages, like `matchers.md` fix up some links to be more useful, like linking inline matcher tokens to their respective type, and matchers to their respective `h3` header.

### DIFF
--- a/src/docs/markdown/caddyfile/directives.md
+++ b/src/docs/markdown/caddyfile/directives.md
@@ -2,9 +2,36 @@
 title: Caddyfile Directives
 ---
 
+<style>
+#directive-table table {
+	margin: 0 auto;
+	overflow: hidden;
+}
+
+#directive-table tr:hover {
+	background: rgba(0, 0, 0, 10%);
+}
+
+#directive-table tr td:first-child {
+	position: relative;
+}
+
+#directive-table a:before {
+	content: '';
+	position: absolute;
+	left: 0;
+	top: 0;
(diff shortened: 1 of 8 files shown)


mholt (approved):
Wow, this is pretty intense, but thanks -- let's give it a shot and see how useful it is!

francislavoie:
FYI @tnolet this was implemented based on your feedback on HN https://news.ycombinator.com/item?id=28245350, please check it out! :grin:

PR #195, opened 2021-11-02
docs: Adding reverse proxy and Let's Encrypt example

miohtama:
You can do both reverse proxy and getting a Let's Encrypt certificate in a single command.

### DIFF
--- a/src/docs/markdown/quick-starts/reverse-proxy.md
+++ b/src/docs/markdown/quick-starts/reverse-proxy.md
@@ -30,6 +30,10 @@ If you don't have permission to bind to low ports, you can proxy from a higher p
 
 Then make a request to [localhost](https://localhost) (or whatever address you specified in `--from`) to see it working!
 
+You can also automatically bind `caddy` reverse proxy to a public IP address and have a Let's Encrypt certificate issued for.
+This command will automatically issue a TLS certificate for `mywebsite.example.com` and start reverse proxy.
+
+<pre><code class="cmd bash">caddy reverse-proxy --from https://mywebsite.example.com --to 127.0.0.1:9000</code></pre>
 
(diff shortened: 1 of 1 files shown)


src/docs/markdown/quick-starts/reverse-proxy.md
```diff
@@ -30,6 +30,10 @@ If you don't have permission to bind to low ports, you can proxy from a higher p
 
 Then make a request to [localhost](https://localhost) (or whatever address you specified in `--from`) to see it working!
 
+You can also automatically bind `caddy` reverse proxy to a public IP address and have a Let's Encrypt certificate issued for.
+This command will automatically issue a TLS certificate for `mywebsite.example.com` and start reverse proxy.
+
+<pre><code class="cmd bash">caddy reverse-proxy --from https://mywebsite.example.com --to 127.0.0.1:9000</code></pre>
```
mholt, line 36:
```suggestion
If you have a public domain name to use, Caddy can serve your proxy over HTTPS using a trusted certificate automatically:

<pre><code class="cmd bash">caddy reverse-proxy --from example.com --to 127.0.0.1:9000</code></pre>
```

mholt (changes requested):
Thanks; I think we can make it a little simpler though.

(And what's up with the change at the end of the file?)

src/docs/markdown/quick-starts/reverse-proxy.md
francislavoie: > (And what's up with the change at the end of the file?)

Some IDEs add or remove the extra newline at the end depending on settings, and it ends up rendering like that in github.

mholt:
Closing due to inactivity

PR #197, opened 2021-11-20
docs: Elaborate on how the `php_fastcgi` directive works

francislavoie:
This is mostly taken from a post I wrote on the forums last month https://caddy.community/t/php-fastcgi-only-serving-php-files-or-wildcard-possible/13918/3 which I thought was a good and thorough explanation of how the directive works, and why it works that way. Did a bit of polish on the wording the fit the language the docs usually use (IMO anyways).

Also added a ToC since the page gets a bit long after these additions, best to give users a way to jump down to the examples quickly.

### DIFF
--- a/src/docs/markdown/caddyfile/directives/php_fastcgi.md
+++ b/src/docs/markdown/caddyfile/directives/php_fastcgi.md
@@ -84,7 +84,6 @@ route {
 
 ### Explanation
 
-
 - The first section deals with canonicalizing the request path. The goal is to ensure that requests that target a directory on disk actually have the trailing slash `/` added to the request path, so that only a single URL is valid for requests to that directory.
 
   This is performed by using a request matcher that matches only requests that _don't_ end in a slash, and which map to a directory on disk which contains an `index.php` file, and if it matches, performs a HTTP 308 redirect with the trailing slash appended. So, for example it would redirect the request to path `/foo` to `/foo/` (appending a `/`, to canonicalize the path to the directory), if `/foo/index.php` exists on disk.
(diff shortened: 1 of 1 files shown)


src/docs/markdown/caddyfile/directives/php_fastcgi.md
```diff
@@ -49,7 +54,7 @@ Since this directive is an opinionated wrapper over a reverse proxy, you can use
 
 ## Expanded form
 
-The `php_fastcgi` directive is the same as the following configuration:
+The `php_fastcgi` directive (with none of the subdirectives configured) is the same as the following configuration. Most modern PHP apps work well with this preset. If yours does not, feel free to borrow from this and customize it as needed instead of using the `php_fastcgi` shortcut.
```
mholt:
```suggestion
The `php_fastcgi` directive (without subdirectives) is the same as the following configuration. Most modern PHP apps work well with this preset. If yours does not, feel free to borrow from this and customize it as needed instead of using the `php_fastcgi` shortcut.
```

src/docs/markdown/caddyfile/directives/php_fastcgi.md
```diff
@@ -77,7 +82,27 @@ route {
 }
 ```
 
-Most modern PHP apps work well with this preset. If yours does not, feel free to borrow from this and customize it as needed instead of using the `php_fastcgi` shortcut.
+### Explanation
+
+Since the above expanded form may be hard to follow, here's an in-depth explanation of what it directive does, and why it works this way.
```
mholt:
```suggestion
```

(Remove fluff, an explanation is great, no need to justify it IMO)

mholt (changes requested):
Thanks for this, people like explanations. Just a minor nit.

mholt (approved):

MestreLion:
That's a great addition!

Could also mention that some in systems, such as Debian and Ubuntu, `php-fpm` by default listens on a _UNIX (file) socket_, not a network address. So the example `php_fastcgi http://127.0.0.1:9000` should be changed to `php_fastcgi unix//run/php/php7.4-fpm.sock` in such systems

PR #198, opened 2021-11-20
docs: Adjustments to global options docs, for clarification

francislavoie:
The main reason for the change is to clarify the usage of `storage`, since it's not obvious how to configure the `file_server` storage module, but I found there's room in many of the options to clarify a bit more their usage and purpose.

Prompted by a topic on the forums https://caddy.community/t/how-to-change-location-of-certificates-in-caddy/14230

### DIFF
--- a/src/docs/markdown/caddyfile/options.md
+++ b/src/docs/markdown/caddyfile/options.md
@@ -106,34 +106,56 @@ Possible options are:
 }
 ```
 
+
 ## General Options
 
 ##### `debug`
-Enables debug mode, which sets all log levels to debug (unless otherwise specified).
+Enables debug mode, which sets the log level to `DEBUG` for the default logger. This may reveal more information about Caddy's internals. Enable this before asking for help on the [community forums](https://caddy.community)!
+
 
 ##### `http_port`
 The port for the server to use for HTTP. For internal use only; does not change the HTTP port for clients. Default: `80`
 
+
 ##### `https_port`
 The port for the server to use for HTTPS. For internal use only; does not change the HTTPS port for clients. Default: `443`
 
+
 ##### `order`
-Sets or changes the standard order of HTTP handler directive(s). Can set directives to be `first` or `last`, or `before` or `after` another directive.
+Sets or changes the [standard order](/docs/caddyfile/directives#directive-order) of HTTP handler directive(s). Can set directives to be ordered `first` or `last`, or inserted `before` or `after` another directive.
+
+For example, to make sure the [`replace-response` plugin's](https://github.com/caddyserver/replace-response) directive is ordered after `encode`, to make sure it manipulates the response body before compression:
+
+```caddy-d
+order replace after encode
+```
+
 
 ##### `storage`
 Configures Caddy's storage mechanism. The default is [`file_system`](/docs/json/storage/file_system/). There are many other available [storage modules](/docs/json/storage/) provided as plugins.
+
+For example, to change the file system's storage location:
+
+```caddy-d
+storage file_system /path/to/custom/location
+```
+
+Changing the storage location is typically done when needing to sync Caddy's storage across multiple instances of Caddy, to make sure they all use the same certificates and keys. See the [Automatic HTTPS section on storage](/docs/automatic-https#storage) for more details.
+
 
 ##### `storage_clean_interval`
 How often to scan storage units for old or expired assets and remove them. These scans exert lots of reads (and list operations) on the storage module, so choose a longer interval for large deployments. The value is a [duration value](/docs/conventions#durations). Default: 24h.
 
 Storage will always be cleaned when the process first starts. Then, a new cleaning will be started this duration after the previous cleaning started if the previous cleaning finished in less than half the time of this interval (otherwise next start will be skipped).
 
+
 ##### `admin`
-Customizes the [admin API endpoint](/docs/api). If `off`, then the admin endpoint will be disabled. If disabled, config changes will be impossible without stopping and starting the server.
+Customizes the [admin API endpoint](/docs/api). If `off`, then the admin endpoint will be disabled. If disabled, config changes will be impossible without stopping and starting the server. Default: `localhost:2019`.
 
 - **origins** configures the list of remotes/origins that are allowed to connect to the endpoint.
 
 - **enforce_origin** enables enforcement of the Origin header. (This is different from enforcing origins generally, which is always done.)
+
 
 ##### `log`
 Customizes the named logger. The name can be passed to indicate a specific logger to customize the behavior for. If no name is specified, the behavior of the default logger is modified. This option can be specified multiple times to configure different loggers. You can read more about the default logger and other logging behaviors in the [logging documentation](/docs/logging).
@@ -146,7 +168,8 @@ Customizes the named logger. The name can be passed to indicate a specific logge
 
 
 ##### `grace_period`
-Defines the grace period for shutting down HTTP servers during config reloads. If clients do not finish their requests within the grace period, the server will be forcefully terminated to allow the reload to complete and free up resources.
+Defines the grace period for shutting down HTTP servers during config reloads. If clients do not finish their requests within the grace period, the server will be forcefully terminated to allow the reload to complete and free up resources. By default, no grace period is set.
+
 
 
 ## TLS Options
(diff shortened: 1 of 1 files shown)


src/docs/markdown/caddyfile/options.md
```diff
@@ -106,35 +106,57 @@ Possible options are:
 }
 ```
 
+
 ## General Options
 
 ##### `debug`
-Enables debug mode, which sets all log levels to debug (unless otherwise specified).
+Enables debug mode, which sets the log level to `DEBUG` for the default logger. This may reveal more information about Caddy's internals. Enable this before asking for help on the [community forums](https://caddy.community)!
```
```suggestion
Enables debug mode, which sets the log level to `DEBUG` for the default logger. This reveals more details that can be useful when troubleshooting (and is very verbose in production). We ask that you enable this before asking for help on the [community forums](https://caddy.community).
```

Slight rewording since "reveal more information about Caddy's internals" sounds a bit like a security leak (which isn't really the case).

src/docs/markdown/caddyfile/options.md
```diff
@@ -106,35 +106,57 @@ Possible options are:
...
 The port for the server to use for HTTP. For internal use only; does not change the HTTP port for clients. Default: `80`
 
+
 ##### `https_port`
 The port for the server to use for HTTPS. For internal use only; does not change the HTTPS port for clients. Default: `443`
 
+
 ##### `order`
-Sets or changes the standard order of HTTP handler directive(s). Can set directives to be `first` or `last`, or `before` or `after` another directive.
+Sets or changes the [standard order](/docs/caddyfile/directives#directive-order) of HTTP handler directive(s). Can set directives to be ordered `first` or `last`, or inserted `before` or `after` another directive.
```
```suggestion
Assigns an order to HTTP handler directive(s). As HTTP handlers execute in a sequential chain, it is necessary for the handlers to be executed in the right order. Standard directives have [a pre-defined order](/docs/caddyfile/directives#directive-order), but if using third-party HTTP handler modules, you'll need to define the order explicitly by either using this option or placing the directive in a [`route` block](/docs/caddyfile/directives/route). Ordering can be described absolutely (`first` or `last`), or relatively (`before` or `after`) to another directive.
```

Not related to your change, but I figure this is clearer / less awkward wording.

src/docs/markdown/caddyfile/options.md
```diff
@@ -106,35 +106,57 @@ Possible options are:
...
+
+For example, to make sure the [`replace-response` plugin's](https://github.com/caddyserver/replace-response) directive is ordered after `encode`, to make sure it manipulates the response body before compression:
```
mholt: This is correct, but it seems like a mistake because it's not clear that the replacement happens on the way back up the handler chain.

```suggestion
For example, to use the [`replace-response` plugin](https://github.com/caddyserver/replace-response), you'd want to make sure its directive is ordered after `encode` so that it can perform replacements before the response is encoded (because responses flow up the handler chain, not down):
```

src/docs/markdown/caddyfile/options.md
```diff
@@ -106,35 +106,57 @@ Possible options are:
...
 ##### `storage`
 Configures Caddy's storage mechanism. The default is [`file_system`](/docs/json/storage/file_system/). There are many other available [storage modules](/docs/json/storage/) provided as plugins.
 
+For example, to change the file system's storage location:
+
+```caddy-d
+storage file_system /path/to/custom/location
+```
+
+Changing the storage location is typically done when needing to sync Caddy's storage across multiple instances of Caddy, to make sure they all use the same certificates and keys. See the [Automatic HTTPS section on storage](/docs/automatic-https#storage) for more details.
```
```suggestion
Customizing the storage module is typically needed when syncing Caddy's storage across multiple instances of Caddy to make sure they all use the same certificates and keys. See the [Automatic HTTPS section on storage](/docs/automatic-https#storage) for more details.
```

src/docs/markdown/caddyfile/options.md
```diff
@@ -106,35 +106,57 @@ Possible options are:
...
+
 ##### `storage_clean_interval`
 How often to scan storage units for old or expired assets and remove them. These scans exert lots of reads (and list operations) on the storage module, so choose a longer interval for large deployments. The value is a [duration value](/docs/conventions#durations). Default: 24h.
 
 Storage will always be cleaned when the process first starts. Then, a new cleaning will be started this duration after the previous cleaning started if the previous cleaning finished in less than half the time of this interval (otherwise next start will be skipped).
 
+
 ##### `admin`
-Customizes the [admin API endpoint](/docs/api). If `off`, then the admin endpoint will be disabled. If disabled, config changes will be impossible without stopping and starting the server.
+Customizes the [admin API endpoint](/docs/api). If `off`, then the admin endpoint will be disabled. If disabled, config changes will be impossible without stopping and starting the server. Default: `localhost:2019`.
```
mholt: Not sure I love documenting the default in two separate places, we already link to the canonical docs page for the API too.
```suggestion
Customizes the [admin API endpoint](/docs/api). If `off`, then the admin endpoint will be disabled. If disabled, config changes will be impossible without stopping and starting the server.
```

src/docs/markdown/caddyfile/options.md
```diff
@@ -146,62 +168,77 @@ Customizes the named logger. The name can be passed to indicate a specific logge
 
 
 ##### `grace_period`
-Defines the grace period for shutting down HTTP servers during config reloads. If clients do not finish their requests within the grace period, the server will be forcefully terminated to allow the reload to complete and free up resources.
+Defines the grace period for shutting down HTTP servers during config reloads. If clients do not finish their requests within the grace period, the server will be forcefully terminated to allow the reload to complete and free up resources. By default, no grace period is set.
```
mholt, line 171:
```suggestion
Defines the grace period for shutting down HTTP servers during config reloads. If clients do not finish their requests within the grace period, the server will be forcefully terminated to allow the reload to complete and free up resources.
```

Same here, not sure how much I love duplicating docs, makes them harder to maintain.

src/docs/markdown/caddyfile/options.md
```diff
@@ -146,62 +168,77 @@ Customizes the named logger. The name can be passed to indicate a specific logge
...
 Configures [On-Demand TLS](/docs/automatic-https#on-demand-tls) where it is enabled, but does not enable it (to enable it, use the [on_demand `tls` subdirective](/docs/caddyfile/directives/tls#syntax)). Highly recommended if using in production environments, to prevent abuse.
 
 - **ask** will cause Caddy to make an HTTP request to the given URL with a query string of `?domain=` containing the value of the domain name. If the endpoint returns 200 OK, Caddy will be authorized to obtain a certificate for that name.
 
 - **interval** and **burst** allows `<n>` certificate operations within `<duration>` interval.
 
+
 ##### `key_type`
-Specifies the type of key to generate for TLS certificates; only change this if you have a specific need to customize it.
+Specifies the type of key to generate for TLS certificates; only change this if you have a specific need to customize it. The possible values are: `ed25519`, `p256`, `p384`, `rsa2048`, `rsa4096`.
```
mholt, line 222: Same; can't we just link to the docs that explain this instead of duplicating?

mholt (changes requested):
Thanks for this, though I have some concerns about duplicating documentation.

src/docs/markdown/caddyfile/options.md
francislavoie: Yeah, struggled with this one a bit. Thanks :+1:

src/docs/markdown/caddyfile/options.md
francislavoie: Everything else documents their defaults though :man_shrugging: I think it's helpful to mention it here in case they land here before `/docs/api`, or don't even go to that page. But point taken. I can go either way on this one.

src/docs/markdown/caddyfile/options.md
mholt: I did too :sweat_smile:

src/docs/markdown/caddyfile/options.md
francislavoie: We're not duplicating in this case, because it's not mentioned otherwise that this _isn't_ turned on by default.

src/docs/markdown/caddyfile/options.md
francislavoie: The only place is in the syntax doc above, but we know users often don't look there :thinking:

mholt: True, how about we link to it again right here, saying something like "Check these docs for possible values".

francislavoie: The syntax docs are kinda huge tho, can't really do a targeted link back up there without making it confusing.

mholt (approved):
Ok, thanks for the improvements, this is a good step forward.

PR #199, opened 2021-11-24
docs: Mention the timeout/buffer defaults for the proxy transports

francislavoie:
Pairs with https://github.com/caddyserver/caddy/pull/4436.

I could split these changes in two, some changes should only land in 2.5.0, but others are pure documentation that could be pushed live now. But 🤷‍♂️ I think it's fine for these to wait until 2.5.0, what's another month or whatever? 😅


mholt (approved):
This is a nice bunch of improvements. Thank you!

2.5 can probably go out soon. I don't have much planned for it...

PR #203, opened 2022-01-08
reverse_proxy: document how to strip a path prefix

simon04:
Documentation for https://github.com/caddyserver/caddy/issues/4512


francislavoie (approved):
Seems maybe a smidge redundant (cause I really would hope users would be smart enough to realize they can skip the rewrite if they don't need to add a new prefix), but sure, why not 👍

PR #208, opened 2022-02-01
Update route.md: fix link

axelsimon:
Hope this helps :)

### DIFF
--- a/src/docs/markdown/caddyfile/directives/route.md
+++ b/src/docs/markdown/caddyfile/directives/route.md
@@ -67,7 +67,7 @@ And now `file_server` will be chained in before `redir` because the order is tak
 
 There are other directives that can wrap HTTP handler directives, but each has its use depending on the behavior you want to convey:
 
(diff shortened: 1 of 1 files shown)


francislavoie:
Hmm, this doesn't seem right. "re-ordered" is not a directive or other kind of code related thing, so wrapping it in backticks to use `<code>` around it doesn't make sense.

francislavoie:
This made me dig deeper to see what's going on, I think the problem is that caddyserver.com is running a build of Caddy which was missing some fixes in the Markdown parser https://github.com/yuin/goldmark which made that link parse incorrectly (I'm guessing because it was preceded by `2)` on the same line which made it incorrectly parse the link which also contains `)` characters).

I'm asking @mholt to update the site with a fixed version.

mholt:
Phew, finally fixed. The site's now running the latest commits of Caddy, and Caddy v2.5 will have the fix.

axelsimon:
Ok, good to hear this eventually helped uncover something!

PR #219, opened 2022-03-16
docs: Fix Debian installation instructions

maxtruxa:
Adding GPG keys to `/etc/apt/trusted.gpg` or `/etc/apt/trusted.gpg.d/` is highly discouraged (this includes usage of `apt-key add`), because these keys are trusted to sign package metadata from **all** sources.
Some references:
- [Debian Wiki article](https://wiki.debian.org/DebianRepository/UseThirdParty) with basic instructions.
- [Cloudflare Blog article](https://blog.cloudflare.com/dont-use-apt-key/) for an in-depth description of the issue.

While it doesn't make the installation instructions any more pretty, this is actually the right way to do it.

Another improvement would be to not download the list file but hardcode the URLs. This would remove that ugly sed call as well.
```sh
curl -1sLf 'https://dl.cloudsmith.io/public/caddy/stable/gpg.key' \
  | gpg --dearmor \
  | sudo tee /usr/share/keyrings/caddy-stable-archive-keyring.gpg >/dev/null
cat - | sudo tee /etc/apt/sources.list.d/caddy-stable.list <<EOS
# Source: Caddy
# Site: https://github.com/caddyserver/caddy
# Repository: Caddy / stable
# Description: Fast, multi-platform web server with automatic HTTPS

deb [signed-by=/usr/share/keyrings/caddy-stable-archive-keyring.gpg] https://dl.cloudsmith.io/public/caddy/stable/deb/debian any-version main
deb-src [signed-by=/usr/share/keyrings/caddy-stable-archive-keyring.gpg] https://dl.cloudsmith.io/public/caddy/stable/deb/debian any-version main
EOS
```

### DIFF
--- a/src/docs/markdown/install.md
+++ b/src/docs/markdown/install.md
@@ -45,16 +45,24 @@ Installing this package automatically starts and runs Caddy as a [systemd servic
 Stable releases:
 
 <pre><code class="cmd"><span class="bash">sudo apt install -y debian-keyring debian-archive-keyring apt-transport-https</span>
-<span class="bash">curl -1sLf 'https://dl.cloudsmith.io/public/caddy/stable/gpg.key' | sudo tee /etc/apt/trusted.gpg.d/caddy-stable.asc</span>
-<span class="bash">curl -1sLf 'https://dl.cloudsmith.io/public/caddy/stable/debian.deb.txt' | sudo tee /etc/apt/sources.list.d/caddy-stable.list</span>
+<span class="bash">curl -1sLf 'https://dl.cloudsmith.io/public/caddy/stable/gpg.key' \\</span>
+<span class="bash-continuation">  | gpg --dearmor \\</span>
+<span class="bash-continuation">  | sudo tee /usr/share/keyrings/caddy-stable-archive-keyring.gpg >/dev/null</span>
+<span class="bash">curl -1sLf 'https://dl.cloudsmith.io/public/caddy/stable/debian.deb.txt' \\</span>
+<span class="bash-continuation">  | sed 's|^deb\(-src\)\?|\0 [signed-by=/usr/share/keyrings/caddy-stable-archive-keyring.gpg]|' \\</span>
+<span class="bash-continuation">  | sudo tee /etc/apt/sources.list.d/caddy-stable.list</span>
 <span class="bash">sudo apt update</span>
 <span class="bash">sudo apt install caddy</span></code></pre>
 
 Testing releases (includes betas and release candidates):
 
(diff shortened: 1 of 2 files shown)


francislavoie:
_Sigh_. This is so annoying. We made similar changes for different reasons in https://github.com/caddyserver/website/pull/177

Looking at https://pkg.cloudflareclient.com/install as an example, I think we could simplify it more:

```
$ curl https://dl.cloudsmith.io/public/caddy/stable/gpg.key | sudo gpg --yes --dearmor --output /usr/share/keyrings/caddy-stable-archive-keyring.gpg
$ echo 'deb [signed-by=/usr/share/keyrings/caddy-stable-archive-keyring.gpg] https://dl.cloudsmith.io/public/caddy/stable/deb/debian any-version main' | sudo tee /etc/apt/sources.list.d/caddy-stable.list
```

WDYT? I don't think we actually need the `deb-src` line, right?

maxtruxa:
While I prefer not running `gpg` as root unnecessarily, it probably doesn't really matter. In theory, we could even skip dearmoring and reference the ASCII-armored key from `signed-by` (supported by `apt (>= 1.4)`, i.e. stretch and later), but IMO it would be unexpected to find ASCII-armored keys in `/usr/share/keyrings`.

For just installing the pre-built package the `deb-src` isn't needed, but I don't know if some people rebuild customized Caddy packages from source. For the basic installation instructions we can omit it I guess.

Alternatively, putting the key download on a single line and just removing the Cloudsmith-specific preamble makes the instructions pretty short as well:
```sh
curl -1sLf 'https://dl.cloudsmith.io/public/caddy/stable/gpg.key' | gpg --dearmor | sudo tee /usr/share/keyrings/caddy-stable-archive-keyring.gpg >/dev/null
cat - | sudo tee /etc/apt/sources.list.d/caddy-stable.list <<EOS
deb [signed-by=/usr/share/keyrings/caddy-stable-archive-keyring.gpg] https://dl.cloudsmith.io/public/caddy/stable/deb/debian any-version main
deb-src [signed-by=/usr/share/keyrings/caddy-stable-archive-keyring.gpg] https://dl.cloudsmith.io/public/caddy/stable/deb/debian any-version main
EOS
```

francislavoie:
Well, crap. Looks like CloudSmith actually made the change to add `signed-by` to the https://dl.cloudsmith.io/public/caddy/stable/debian.deb.txt and I wasn't aware, so new installs are breaking because it looks for the GPG key in the wrong place.

See https://github.com/caddyserver/dist/issues/83

I'll do a quick fixup to this PR and merge it -- we don't need to change the sources command, but we do need to update the key installation part.

PR #225, opened 2022-04-07
Elaborate on log formats, proxy headers and examples

francislavoie:
Logging docs were lacking the exact supported formats for time/duration/level. We were also missing Caddyfile support for `duration_format` altogether, so I added that here https://github.com/caddyserver/caddy/pull/4684

Proxy docs for headers showed the `-` and `+` syntax at the top, but didn't mention how to use it anywhere. Wrote out some quick representative examples too. This part could use a bit of polish maybe, not sure about the fields/values I used etc.

And a couple typo fixes elsewhere.


francislavoie:
Looks like we forgot HTTP transport resolvers config https://github.com/caddyserver/caddy/issues/4685, added

mholt (approved):
LGTM. Thanks for all the constant little improvements to the docs!

PR #230, opened 2022-05-11
css: Rework how paper is positioned

francislavoie:
On Firefox, we were seeing an extra large horizontal scrollbar. Didn't happen on Chrome though. I think it's an edgecase bug with Firefox... but I think we can take a better approach that can avoid it anyways.

The CSS for `paper` classes was set up so that they would stack/overlap onto eachother. It was using `left: -100%;` and `left: -200%;` on `paper2` and `paper3`, to stack the next two on top of the first one, all three being relative positioned.

The problem comes because I think Firefox calculates the scrollbar width before relative positioning calculations come into play, so because the elements were rendered way off to the right before being moved over, it would "grow" the page really wide. Go to https://caddyserver.com/docs/ on any browser, element inspector, find `#paper2` and `.paper3`, uncheck the `left` CSS prop on both of those. You'll see that they render over to the right.

An alternate way to do the same thing is to use negative margins to stack the elements. I think margins are calculated earlier, so on Firefox this doesn't cause a scrollbar problem. Tested on Chrome too, looks right this way too.

I did need to use a `calc()` unfortunately to get it pixel-perfect positioning compared to how it was before, because `#paper1` has `margin-left: 20px` and then `#paper2` inherits that margin so it needs to remove the extra 20px to realign. `calc()` is cheap though. It's fine.

Also, need `z-index: -1;` to make sure the BG papers go behind the page content.

### DIFF
--- a/src/resources/css/docs.css
+++ b/src/resources/css/docs.css
@@ -182,8 +182,6 @@ main nav li li a {
 }
 
 .paper {
-	position: relative;
-	flex-shrink: 0;
 	width: 100%;
 
 	background: white;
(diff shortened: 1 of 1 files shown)


mholt (approved):
Thanks, I'm usually really picky about stuff like this but I guess I assumed Chrome and Firefox had feature parity with basic stuff like positioning these days. Sigh. (And I dev'ed on Chrome at the time.)

francislavoie:
FWIW I don't think it was an issue on Firefox for a long time, but it did start being an issue at least some months ago. I just assumed it was a temp browser bug or w/e, didn't really think about it much. Looked into it with element inspector at least a couple times and was baffled, didn't make noise about it. But yeah, this seems like a better approach anyways, instead of abusing `flex` and `position: relative`.

PR #231, opened 2022-05-11
docsearch: Upgrade to v3

francislavoie:
See:
- https://docsearch.algolia.com/docs/DocSearch-v3
- https://docsearch.algolia.com/docs/migrating-from-v2


mholt (approved):
Thank you, I didn't know about this either. Let's hope it's an improvement :)

PR #232, opened 2022-05-11
logos: In dark mode, use invert + hue-rotate

francislavoie:
The ZeroSSL logo and the bottom Caddy logo didn't look good in dark mode.

The lock logo at the bottom now looks slightly funky but I think it kinda works 😂 

The clever thing here is we invert the colors, but inverting also inverts the colored parts, not only the greyscale; so then you can hue-rotate to un-invert the non-grayscale parts. Pretty neat trick.

---

Before:

image

image

After:

image

image

### DIFF
--- a/src/resources/css/docs.css
+++ b/src/resources/css/docs.css
@@ -935,8 +935,13 @@ td code {
 	}
 
 	#logo {
-		/* TODO: Add some color */
-		filter: invert(1) opacity(.35);
+		filter: invert(1);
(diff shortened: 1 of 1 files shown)


mholt (approved):
Oooooo, shiny. Thanks! That's a neat trick, I didn't know before!

francislavoie:
I learned it from digging into how https://darkreader.org/ works

PR #233, opened 2022-05-12
json: Make URL #fragment jump to the right spot after loading

francislavoie:
I think this is a big usability fix. Click on https://caddyserver.com/docs/json/apps/http/#servers/tls_connection_policies/client_authentication for example; you'd expect this would scroll the browser down to that specific config option, but it doesn't, it just scrolls down _a bit_ at page load, then since the content is loaded async with JS, it doesn't go to the right spot.

We can actually just reset the `window.location.hash` real quick right after the content loads, and that'll trigger the browser scrolling to the right spot. Super nice!

This way, we can actually link to specific spots in the JSON docs and it'll actually bring the user there.


mholt (approved):
Thanks, good idea!

PR #240, opened 2022-06-05
A few tips and pointers for newcomers

phish108:
I tried to make a few things clearer with respect to the misconceptions I had.

I used the markdown="1" attribute in the aside-tags I added. Most markdown preprocessors accept this attribute for nesting markdown in HTML tags. You did not use this attribute, so I am unsure, if this works on your website.

### DIFF
--- a/src/docs/markdown/getting-started.md
+++ b/src/docs/markdown/getting-started.md
@@ -56,6 +56,7 @@ This blocks forever, but what is it doing? At the moment... nothing. By default,
 
 We can make Caddy useful by giving it a config. This can be done many ways, but we'll start by making a POST request to the [/load](/docs/api#post-load) endpoint using `curl` in the next section.
 
+
 ## Your first config
 
 To prepare our request, we need to make a config. At its core, Caddy's configuration is simply a [JSON document](/docs/json/).
@@ -124,11 +125,12 @@ Another way to configure Caddy is with the [**Caddyfile**](/docs/caddyfile). The
 respond "Hello, world!"
 ```
 
+
 Save that to a file named `Caddyfile` (no extension) in the current directory.
 
 <aside class="complete">Make a Caddyfile</aside>
 
-Next we check if the configuration is valid. 
+Next we check if our new configuration is valid. 
 
 Stop Caddy if it is already running (Ctrl+C), then run:
 
@@ -198,7 +200,7 @@ Kind of boring | Kind of fun
 You will need to decide which is best for your use case.
 
 <aside class="tip" markdown="1">
-	The Caddyfile offers shortcuts for many common tasks that are require complex JSON configurations. Use `caddy adapt` to learn how these shortcuts work in JSON.
+	The Caddyfile offers shortcuts for many common tasks that require quite complex JSON configurations. You can use `caddy adapt` to learn how these shortcuts are translated between the formats.
 </aside>
 
 It is important to note that both JSON and the Caddyfile (and [any other supported config adapter](/docs/config-adapters)) can be used with [Caddy's API](/docs/api). However, you only get the full range of Caddy's functionality and API features if you use JSON. If using a config adapter, the only way to load or change the config with the API is the [/load endpoint](/docs/api#post-load).
@@ -233,6 +235,7 @@ But most people will use JSON+API or Caddyfile+CLI combinations.
 As you can see, Caddy is well-suited for a wide variety of use cases and deployments!
 
 <aside class="complete">Compare API and config files</aside>
+
 
 
 ## Start, stop, run
--- a/src/docs/markdown/getting-started.md
+++ b/src/docs/markdown/getting-started.md
@@ -34,6 +34,10 @@ Let's start by running it:
 
 Oops; without a subcommand, the `caddy` command only displays help text. You can use this any time you forget what to do.
 
+<aside class="tip" markdown="1">
+	Caddy is controlled through the `caddy` command. It offers many useful tools for different tasks while running Caddy. Have a closer look at [ `caddy`'s subcommands ](https://caddyserver.com/docs/command-line) to learn what you can do. 
+</aside>
+
 To start Caddy as a daemon, use the `run` subcommand:
 
 <pre><code class="cmd bash">caddy run</code></pre>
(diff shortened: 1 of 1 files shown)


src/docs/markdown/getting-started.md
```diff
@@ -34,6 +34,10 @@ Let's start by running it:
 
 Oops; without a subcommand, the `caddy` command only displays help text. You can use this any time you forget what to do.
 
+<aside class="tip" markdown="1">
+	Caddy is controlled through the `caddy` command. It offers many useful tools for different tasks while running Caddy. Have a closer look at [ `caddy`'s subcommands ](https://caddyserver.com/docs/command-line) to learn what you can do. 
+</aside>
```
francislavoie, line 39: Unfortunately, that doesn't work 😢 

image

Might be a good idea to request support for this from https://github.com/yuin/goldmark which is what Caddy uses

francislavoie: There's some mention of `markdown=1` in the CommonMark spec FYI https://spec.commonmark.org/0.30/#example-189 but I'll leave it up to you to request support for this

francislavoie: Oh wow, this works, actually:

```suggestion
<aside class="tip">

Caddy is controlled through the `caddy` command. It offers many useful tools for different tasks while running Caddy. Have a closer look at [ `caddy`'s subcommands ](https://caddyserver.com/docs/command-line) to learn what you can do. 

</aside>
```

Noticed this from reading that CommonMark section, it mentions using the newlines as an alternative.

image

I'd suggest adding some extra newlines before and after the `<aside>` to give it room to breathe around the other paragraphs.

mholt: I think I'm surprised that works, although maybe I do remember seeing that it works except something was off, maybe it was the line breaks or something.

src/docs/markdown/getting-started.md
```diff
@@ -127,6 +130,8 @@ Save that to a file named `Caddyfile` (no extension) in the current directory.
 
 <aside class="complete">Make a Caddyfile</aside>
 
+Next we check if our new configuration is valid. 
+
```
mholt, line 134: This isn't really true. We're checking if our configuration is _syntactically_ valid, but the point of `caddy adapt` (without `--validate`) is not to check for syntactic validity, it's to convert the config to JSON. I don't think I want users getting the idea that `caddy adapt` is only for validating config (which it doesn't even really do unless you specify `--validate` -- it only checks syntax, and that's a side-effect, not the objective).

src/docs/markdown/getting-started.md
francislavoie: Paragraphs seem to work fine

image


```md
<aside class="tip">

Caddy is controlled through the `caddy` command. It offers many useful tools for different tasks while running Caddy. Have a closer look at [ `caddy`'s subcommands ](https://caddyserver.com/docs/command-line) to learn what you can do. 

A new line for testing

Another new line for testing

</aside>
```

src/docs/markdown/getting-started.md
```diff
@@ -176,6 +186,7 @@ Full range of Caddy functionality | Most common parts of Caddy functionality
 Easy to generate | Easy to craft by hand
 Easily programmable | Difficult to automate
 Extremely expressive | Moderately expressive
+Hard to read | Easy to read
```
mholt, line 189: So, I disagree on this point. I do think the Caddyfile is easy to read, but I also think the JSON is easy to read. It's definitely more cognitive to _write_ JSON by hand, but not harder to read.

I think we need to keep this table as objective as possible, not subjective, so let's remove this line.

src/docs/markdown/getting-started.md
```diff
@@ -135,9 +140,14 @@ Or if you stored the Caddyfile somewhere else or named it something other than `
 
 <pre><code class="cmd bash">caddy adapt --config /path/to/Caddyfile</code></pre>
 
-You will see JSON output! What happened here?
+If you made no errors, you will see JSON output! What happened here?
+
+We just used a [_config adapter_](/docs/config-adapters) to convert our Caddyfile to Caddy's native JSON structure. 
+
+<aside class="tip" markdown="1">
+	It is always good to test the changes in your `Caddyfile` using `caddy adapt`. If there are errors in your `Caddyfile`, you will see where you made a mistake instead of the JSON.
```
mholt, line 148: > It is always good to test the changes in your `Caddyfile` using `caddy adapt`.

Most users won't need to do this and this step is actually redundant for a lot of workflows (including my own). Caddy will reject syntactically invalid configs regardless of format.

As above, I think this tip should be removed since it gives the impression that `caddy adapt` is for config validation, but it's not. That's what [`caddy validate`](https://caddyserver.com/docs/command-line#caddy-validate) is for.

src/docs/markdown/getting-started.md
```diff
@@ -188,7 +199,11 @@ Kind of boring | Kind of fun
 
 You will need to decide which is best for your use case.
 
-It is important to note that both JSON and the Caddyfile (and [any other supported config adapter](/docs/config-adapters)) can be used with [Caddy's API](/docs/api). However, you get the full range of Caddy's functionality and API features if you use JSON. If using a config adapter, the only way to load or change the config with the API is the [/load endpoint](/docs/api#post-load).
+<aside class="tip" markdown="1">
+	The Caddyfile offers shortcuts for many common tasks that require quite complex JSON configurations. You can use `caddy adapt` to learn how these shortcuts are translated between the formats.
```
mholt, line 203: Hm, maybe something like this a little more to the point:

```suggestion
	Use `caddy adapt` often to learn one way to represent your Caddyfile as native JSON. (The output of `caddy adapt` is not always optimal or the most concise, and the best JSON configs are designed by humans, not a config adapter.)
```

Because there are configs that are more concise in JSON than in Caddyfile.

mholt (changes requested):
Thanks for the suggestions. I've left some feedback here, and I think we can work with some of this.

src/docs/markdown/getting-started.md
phish108: OK. 

I was thinking as readability in terms of understanding what is going on.

mholt:
Closing due to inactivity

PR #237, opened 2022-05-27
reverse_proxy: Document header policy fallback

otbutz:
As discussed with @francislavoie in https://github.com/caddyserver/caddy/issues/4809#issuecomment-1139421336

PR #241, opened 2022-06-05
docs: Some random polish

francislavoie:
Just a bunch of little things that I think would help users from frequently asked forum questions.

Small followup to #237, I wanted to make sure the rest of the selection policies were accurately described.

### DIFF
--- a/src/docs/markdown/caddyfile/concepts.md
+++ b/src/docs/markdown/caddyfile/concepts.md
@@ -328,7 +328,7 @@ A default value can be specified for when the environment variable is not found,
 {$DOMAIN:localhost}
 ```
 
-If you want to defer the substitution of an environment variable until runtime, you can use the [standard `{env.*}` placeholders](/docs/conventions#placeholders). Note that not all config parameters support these placeholders though, since module developers need to add a line of code to perform the replacement; placeholder replacement is not automatic and needs to be performed at runtime. If it doesn't seem to work, please file an issue to request support for it.
+If you want to defer the substitution of an environment variable until runtime, you can use the [standard `{env.*}` placeholders](/docs/conventions#placeholders). Note that not all config parameters support these placeholders though, since module developers need to add a line of code to perform the replacement. If it doesn't seem to work, please file an issue to request support for it.
 
 
 
--- a/src/docs/markdown/caddyfile/concepts.md
+++ b/src/docs/markdown/caddyfile/concepts.md
@@ -328,7 +328,7 @@ A default value can be specified for when the environment variable is not found,
 {$DOMAIN:localhost}
 ```
 
-If you want to defer the substitution of an environment variable until runtime, you can use the [standard `{env.*}` placeholders](/docs/conventions#placeholders). Note that not all parts of the config support these placeholders though, since a line of code is necessary to perform the replacement; placeholder replacement is not automatic and needs to be performed at runtime. If it doesn't seem to work, please file an issue to request support for it.
+If you want to defer the substitution of an environment variable until runtime, you can use the [standard `{env.*}` placeholders](/docs/conventions#placeholders). Note that not all config parameters support these placeholders though, since module developers need to add a line of code to perform the replacement; placeholder replacement is not automatic and needs to be performed at runtime. If it doesn't seem to work, please file an issue to request support for it.
 
 
 
--- a/src/docs/markdown/caddyfile/concepts.md
+++ b/src/docs/markdown/caddyfile/concepts.md
@@ -24,9 +24,9 @@ The Caddyfile's structure can be described visually:
 
 Key points:
 
-- An optional **global options block** can be the very first thing in the file.
-- Otherwise, the first line of the Caddyfile is **always** the address(es) of the site to serve.
-- All directives and matchers **must** go in a site block. There is no global scope or inheritence across site blocks.
+- An optional [**global options block**](#global-options) can be the very first thing in the file.
+- Otherwise, the first line of the Caddyfile is **always** the [address(es)](#addresses) of the site to serve.
+- All [directives](#directives) and [matchers](#matchers) **must** go in a site block. There is no global scope or inheritence across site blocks.
 - If there is **only one site block**, its curly braces `{ }` are optional.
 
 A Caddyfile consists of at least one or more site blocks, which always starts with one or more [addresses](#addresses) for the site. Any directives appearing before the address will be confusing to the parser.
@@ -145,7 +145,7 @@ Inside quoted tokens, all other characters are treated literally, including spac
 
 ```caddy-d
 directive "first line
-second line"
+	second line"
 ```
 
 You can also use a backtick <code>`</code> to quote tokens; these are convenient when tokens themselves contain double quotes, e.g. JSON text:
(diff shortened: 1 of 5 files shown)


src/docs/markdown/caddyfile/concepts.md
```diff
@@ -145,7 +145,7 @@ Inside quoted tokens, all other characters are treated literally, including spac
 
 ```caddy-d
 directive "first line
-second line"
+	second line"
```
mholt, line 148: Is this change just to illustrate that tabs are taken literally inside quotes?

src/docs/markdown/caddyfile/concepts.md
```diff
@@ -319,15 +320,16 @@ If your configuration relies on environment variables, you can use them in the C
...
+Environment variables in this form are substituted before Caddyfile parsing begins, so they can expand to empty values, partial tokens, complete tokens, or even multiple tokens and lines.
 
 A default value can be specified for when the environment variable is not found, by using `:` as the delimiter between the variable name and the default value:
 
 ```caddy
 {$DOMAIN:localhost}
 ```
 
-If you want to defer the substitution of an environment variable until runtime, you can use the [standard `{env.*}` placeholders](/docs/conventions#placeholders).
+If you want to defer the substitution of an environment variable until runtime, you can use the [standard `{env.*}` placeholders](/docs/conventions#placeholders). Note that not all parts of the config support these placeholders though, since a line of code is necessary to perform the replacement; placeholder replacement is not automatic and needs to be performed at runtime. If it doesn't seem to work, please file an issue to request support for it.
```
mholt: > since a line of code is necessary to perform the replacement

I know we talked about this earlier in Slack, and I don't remember the full conversation now, but my reading of it tonight makes me think some users will assume this means they need to write a line of code to make placeholders work. I don't want to give that impression, so maybe this instead?

```suggestion
If you want to defer the substitution of an environment variable until runtime, you can use the [standard `{env.*}` placeholders](/docs/conventions#placeholders). Note that not all config parameters support placeholders yet, but most common ones do. If it doesn't seem to work for you, please file an issue to request support for it.
```

src/docs/markdown/caddyfile/directives/reverse_proxy.md
```diff
@@ -242,8 +242,8 @@ Active health checks perform health checking in the background on a timer:
 
 Passive health checks happen inline with actual proxied requests:
 
-- **fail_duration** <span id="fail_duration"/>  is a [duration value](/docs/conventions#durations) that defines how long to remember a failed request. A duration > 0 enables passive health checking.
-- **max_fails** <span id="max_fails"/> is the maximum number of failed requests within `fail_duration` that are needed before considering a backend to be down; must be >= 1; default is 1.
+- **fail_duration** <span id="fail_duration"/>  is a [duration value](/docs/conventions#durations) that defines how long to remember a failed request. A duration > `0` enables passive health checking; it's off by default. A reasonable starting point might be `30s`, to balance error rates with responsiveness when bringing an unhealthy upstream back online; but feel free to experiment to find the right balance for your usecase.
```
mholt:
```suggestion
- **fail_duration** <span id="fail_duration"/>  is a [duration value](/docs/conventions#durations) that defines how long to remember a failed request. A duration > `0` enables passive health checking; the default is  (off). A reasonable starting point might be `30s` to balance error rates with responsiveness when bringing an unhealthy upstream back online; but feel free to experiment to find the right balance for your usecase.
```

src/docs/markdown/caddyfile/patterns.md
```diff
@@ -194,6 +195,8 @@ If your SPA is coupled with an API or other server-side-only endpoints, you will
 
 ```caddy
 example.com {
+	encode gzip
```
mholt, line 198: Maybe a comment explaining that this is optional? I really, really try to keep examples the bare minimum so that readers know what is relevant, since that's the point of the example.

mholt (changes requested):
Thanks, these look pretty good overall. Just a few thoughts from my reading through it (after midnight, so YMMV).

src/docs/markdown/caddyfile/concepts.md
francislavoie: Yeah, to show whitespace is preserved/allowed

src/docs/markdown/caddyfile/concepts.md
francislavoie: Yeah but I _really_ want to imply somehow that "plugin/module authors need to implement it, otherwise it won't work". It's been a repeated thing where a plugin author didn't add support and we get questions about it on the forums.

src/docs/markdown/caddyfile/patterns.md
francislavoie: Well it is relevant. There's never a situation where you wouldn't want this on, when you're serving an SPA. It would be terrible if they didn't have compression on, cause then all their JS/CSS/HTML bandwidth would suffer.

src/docs/markdown/caddyfile/directives/reverse_proxy.md
francislavoie: "(off)" reads weird to me 🤔

I agree with the comma change though

francislavoie: Did you mean to write ```default is `0` (off)``` ?

src/docs/markdown/caddyfile/concepts.md
francislavoie: Made some adjustments here, hopefully it's better.

src/docs/markdown/caddyfile/directives/reverse_proxy.md
mholt: Yes, I did! Oops

This is what happens when Matt does code review after midnight

src/docs/markdown/caddyfile/concepts.md
```diff
@@ -319,15 +320,16 @@ If your configuration relies on environment variables, you can use them in the C
...
+Environment variables in this form are substituted before Caddyfile parsing begins, so they can expand to empty values, partial tokens, complete tokens, or even multiple tokens and lines.
 
 A default value can be specified for when the environment variable is not found, by using `:` as the delimiter between the variable name and the default value:
 
 ```caddy
 {$DOMAIN:localhost}
 ```
 
-If you want to defer the substitution of an environment variable until runtime, you can use the [standard `{env.*}` placeholders](/docs/conventions#placeholders).
+If you want to defer the substitution of an environment variable until runtime, you can use the [standard `{env.*}` placeholders](/docs/conventions#placeholders). Note that not all config parameters support these placeholders though, since module developers need to add a line of code to perform the replacement; placeholder replacement is not automatic and needs to be performed at runtime. If it doesn't seem to work, please file an issue to request support for it.
```
mholt:
```suggestion
If you want to defer the substitution of an environment variable until runtime, you can use the [standard `{env.*}` placeholders](/docs/conventions#placeholders). Note that not all config parameters support these placeholders though, since module developers need to add a line of code to perform the replacement. If it doesn't seem to work, please file an issue to request support for it.
```

I think we can remove that one bit, as that makes more sense if the audience is a module developer. I believe this conveys the idea well enough without being confusing.

mholt (changes requested):
Thanks for the revisions -- looks mostly good now, though I think the bit about placeholders can still be cleared up a little.

mholt (approved):
Very nice! Thank you!

PR #247, opened 2022-07-08
2.5.2 docs

mholt:
Let's try this again :laughing:


src/docs/markdown/api.md
```diff
@@ -237,6 +243,26 @@ but with an ID, the path becomes
 
 which is much easier to remember and write by hand.
 
+## Concurrent config changes
+
+<aside class="tip">
+	This section is for all <code>/config/</code> endpoints. It is experimental and subject to change.
+</aside>
```
francislavoie:
```md
<aside class="tip">

This section is for all `/config/` endpoints. It is experimental and subject to change.

</aside>
```

If you use newlines you can use markdown inside the aside

src/docs/markdown/api.md
```diff
@@ -289,4 +315,21 @@ Each entry in the JSON array is a configured [upstream](/docs/json/apps/http/ser
...
+If your goal is to determine a backend's availability, you will need to cross-check relevant properties of the upstream against the handler configuration you are utilizing. For example, if you've enabled [passive health checks](/docs/json/apps/http/servers/routes/handle/reverse_proxy/health_checks/passive/) for your proxies, then you need to also take into consideration the `fails` and `num_requests` values to determine if an upstream is considered available: check that the `fails` amount is less than your configured maximum amount of failures for your proxy (i.e. [`max_fails`](/docs/json/apps/http/servers/routes/handle/reverse_proxy/health_checks/passive/max_fails/)), and that `num_requests` is less than or equal to your configured amount of maximum requests per upstream (i.e. [`unhealthy_request_count`](/docs/json/apps/http/servers/routes/handle/reverse_proxy/health_checks/passive/unhealthy_request_count/) for the whole proxy, or [`max_requests`](/docs/json/apps/http/servers/routes/handle/reverse_proxy/upstreams/max_requests/) for individual upstreams).
+
+
+## POST /adapt
```
francislavoie: I think this should go earlier than the ca/reverse_proxy ones since those are kinda plugins, but adapt is a built-in of the admin module

src/docs/markdown/api.md
```diff
@@ -246,7 +246,9 @@ which is much easier to remember and write by hand.
 ## Concurrent config changes
 
 <aside class="tip">
-	This section is for all <code>/config/</code> endpoints. It is experimental and subject to change.
+
+	This section is for all `/config/` endpoints. It is experimental and subject to change.
```
francislavoie: I think you need to remove the leading tab too otherwise it gets parsed as a code block; needs to be gutter aligned as markdown

mholt: Ah... that's annoying. I might just keep it as HTML. It's not that ugly in this case. Thanks

francislavoie: Hard disagree, it's sooo much better because we can do more complex stuff in there. Just put extra newlines before/after the `<aside>` and it reads ok IMO

mholt:
Thanks for your help Francis!

PR #249, opened 2022-07-20
download: add warning message for macOS users to unquarantine the binary

mohammed90:
Recent releases of macOS automatically quarantine unsigned binaries downloaded from the internet. The internet is full of workarounds, including disabling this protection completely for the entire OS, which we don't want the users to stumble upon. The better workaround is to remove the quarantine bit using the command `xattr -d com.apple.quarantine`.

Given I had to swim through numerous pages to find this proper workaround, it seems reasonable we include a warning for the users on how to un-quarantine the binary to avoid them implementing the worse alternative of disabling the protection for the entire OS.

### DIFF
--- a/src/download.html
+++ b/src/download.html
@@ -74,6 +74,11 @@
 			<div class="text-center">
 				<span class="warning">⚠️ Only choose plugins you need and trust</span>
 			</div>
+			<div class="text-center" id="darwin-warning">
+				<span class="warning">⚠️ Run the following against the downloaded binary:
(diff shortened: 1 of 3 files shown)


mholt (changes requested):
This is a great tip. Can we use CSS to make it `display: none` by default?

mohammed90:
> This is a great tip. Can we use CSS to make it `display: none` by default?

Done

mholt (approved):

PR #252, opened 2022-08-17
remove extra word

danfarinoeyecue:


francislavoie:
Hah, oops, I probably started a sentence but didn't finish the thought and left that there. Thanks!

PR #254, opened 2022-08-24
caddyfile: document ocsp interval global option

benburkert:
Document the `ocsp_interval` option added in caddyserver/caddy#4980.


mholt (approved):
Great, thank you!

PR #255, opened 2022-08-26
HTTP/3 is specified in RFC 9114

appliedprivacy:


mholt:
Thanks for the PR! We are changing all this for Caddy 2.6 soon as http/3 will be on by default and no longer experimental.

I'm going to close this in favor of changes being made locally already, in prep for HTTP/3 being enabled by default and the removal of this option. Thanks for understanding!

PR #259, opened 2022-09-04
install: update webi commands

fu-sen:
webi has adopted new domains `webi.sh` and `webi.ms` and shortened commands.


francislavoie:
Thanks!

PR #264, opened 2022-09-14
docs: Attempt to clarify how handle directives may be re-ordered

asb:
As discussed in <https://github.com/caddyserver/caddy/issues/5037>.

### DIFF
--- a/src/docs/markdown/caddyfile/directives/handle.md
+++ b/src/docs/markdown/caddyfile/directives/handle.md
@@ -6,7 +6,7 @@ title: handle (Caddyfile directive)
 
 Evaluates a group of directives mutually exclusively from other `handle` blocks at the same level of nesting.
 
-The `handle` directive is kind of similar to the `location` directive from nginx config: the first matching `handle` block will be evaluated. Handle blocks can be nested if needed. Only HTTP handler directives can be used inside handle blocks.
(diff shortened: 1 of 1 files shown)


src/docs/markdown/caddyfile/directives/handle.md
```diff
@@ -6,7 +6,7 @@ title: handle (Caddyfile directive)
 
 Evaluates a group of directives mutually exclusively from other `handle` blocks at the same level of nesting.
 
-The `handle` directive is kind of similar to the `location` directive from nginx config: the first matching `handle` block will be evaluated. Handle blocks can be nested if needed. Only HTTP handler directives can be used inside handle blocks.
+The `handle` directive is kind of similar to the `location` directive from nginx config: the first matching `handle` block will be evaluated. `handle` directives at the same level of nesting will be tried in the order they're written in the `Caddyfile`, except if there is a single path matcher, which orders them by the path length. Handle blocks can be nested if needed. Only HTTP handler directives can be used inside handle blocks.
```
mholt:
```suggestion
The `handle` directive is kind of similar to the `location` directive from nginx config: the first matching `handle` block will be evaluated. `handle` directives at the same level of nesting will be tried in the order they're written in the `Caddyfile`, except if there is a single path matcher, which orders them by longest (most specific) path pattern first. Handle blocks can be nested if needed. Only HTTP handler directives can be used inside handle blocks.
```

mholt (changes requested):
Thanks! This is a nice improvement. What do you think about my suggestion?

mholt (approved):

francislavoie:
It's not actually only if there's a single path matcher iirc, doesn't it still apply to named matchers with multiple paths too?

mholt:
@francislavoie With my recent patch, it now applies only to any form of matcher as long as it has just 1 path.

PR #266, opened 2022-09-16
docs: New "Introduction" tutorial

mholt:
[Our current Getting Started tutorial](https://caddyserver.com/docs/getting-started) (as of early September 2022) is a classic guide on our Caddy 2 site. It is required reading, as it demonstrates the basic `caddy` commands, how to load configuration both as native JSON and config adapters, and helps you choose how to run and configure Caddy. It sets a crucial foundation, then lets you choose how to take it from there.

I really like that guide and it needs to stay (in some form or another). But we get a lot of feedback that it's not very _practical_. It doesn't help them actually "get started" with the web server. There's too many directions to take it once you know how to load config and such. But I think that's OK. You just gotta write your config and then apply what you learned in _Getting Started_. It, along with our other quick-starts and tutorials, are excellent universal guides for using Caddy, IMO.

In this PR, I've added a new "Introduction" tutorial to be more practical, inspired by https://ellen.dev/serve-static-site-using-caddy.html. It's probably more comprehensive for beginners. Unfortunately, it steps outside the scope of Caddy and goes into sysadmin territory, which I don't like doing, because it's not our job to document Linux or systemd.

Another problem is that this revision in its current form is Linux-only. That _is_ the vast majority of users by far, but it doesn't represent everyone, especially many who are just trying it "at home" (Windows and Mac). This tutorial should be universal, and I don't want to maintain 3 versions of it, or multiple versions of sub-sections, for various OSes, frankly. I still think it is reasonable to expect that readers know how to use their computers already. To learn how to use the computer is outside the scope of our documentation and is a better fit for our wiki.

Additionally, this tutorial is harder. There's more to read, it moves more slowly, and it feels more complex. The problem is that this reflects poorly on Caddy even though the slowness and complexity is entirely due to how Linux and systemd work.

Lots of people are suggesting using Docker for this universal tutorial, but I have a hard reservation _against_ that, as Docker is often the cause of many complications and confusion in our experience. I think Docker would be fine for an advanced tutorial, but I maintain that the point of our docs is to teach Caddy, not sysadmin.

There is currently not a great transition from _Introduction_ to _Getting Started_.

Still a WIP.

---

**If you're new here and want to read this locally, it's easy:**

```
$ git clone https://github.com/caddyserver/website.git
$ cd website
$ git checkout origin/introduction
$ caddy run
```

then open `https://localhost`

### DIFF
--- a/src/docs/markdown/introduction.md
+++ b/src/docs/markdown/introduction.md
@@ -1,0 +1,276 @@
+---
+title: "Introduction"
+---
+
+# Introduction
+
+This guide will introduce you to setting up a site with Caddy on **Linux** in a matter of minutes, but it is _not the only way_ to do it. There are many ways you can [download](/download), [install](/docs/install), [configure](/docs/introduction), and [run](/docs/running) Caddy. **If you are already comfortable with setting up services using systemd, jump to [Introduction](/docs/introduction).**
+
+You can jump out of this tutorial at any time when you feel like you know what to do next.
+
+**Objectives:**
+- 🔲 Install Caddy as a service
+- 🔲 Discover the unit configuration
+- 🔲 Prepare your site
+- 🔲 Serve your site over HTTPS
+- 🔲 Add a reverse proxy
+- 🔲 Learn how to troubleshoot problems
+
+**Prerequisites:**
+- A computer where you have administrator, root, or sudo privileges
+- Know how to use a terminal / command line
+- Familiarity with Unix permissions
+- Be comfortable editing text files
+- A registered domain name
+
+---
+
+First, ensure no other web servers are running on your machine (to avoid port-binding conflicts).
+
+**[Install Caddy](/docs/install) by following the instructions for your system.** For example, if you're on Ubuntu, follow the steps that use `apt`; on Fedora, use `dnf`; etc. If a package isn't available for your distro, you can also [manually install Caddy as a service](/docs/running) on any Linux machine that has systemd.
+
+<aside class="tip">
+
+All you really need to run Caddy is the executable file itself. "Installing" Caddy could be defined simply as putting Caddy in your PATH. But installing Caddy _as a service_ is best practice for production systems because generally, a service keeps the process running after reboots, implements a tighter permissions/security model, and centralizes logging.
+
+</aside>
+
+Verify that the Caddy service is running:
+
+<pre><code class="cmd bash">systemctl status caddy</code></pre>
+
+You should see output like this:
+
+```plain
+● caddy.service - Caddy
+     Loaded: loaded (/lib/systemd/system/caddy.service; enabled; vendor preset: enabled)
+     Active: active (running) since Tue 2022-09-06 21:15:31 MDT; 1 day 1h ago
+...
+```
+
+Ensure it says "enabled" and "active (running)" -- these are crucial for production. (Enabled means the service will be started automatically after a reboot.)
+
+If Caddy failed to start, the most likely cause is you have another web server running. The output will also show the most recent log lines which you can inspect for error messages. After resolving the issue, run `sudo systemctl start caddy` to try again.
+
+<aside class="complete">Install Caddy as a service</aside>
+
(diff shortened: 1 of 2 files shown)


Plorenzo:
I've been following Caddy and you for a long time now so I think I have a pretty good sense of how reluctant you are _to teach Caddy users how to use computers_  so first of all thanks for giving this a try! 

While in theory and in the ideal world, I agree with you, system administration is outside the scope of Caddy, in practice most users (me a few years ago included, [you can see me struggle in the forum ](https://caddy.community/t/unable-to-run-caddy-as-a-daemon/2364/20)😂) want a default, recommended by the vendor way of deploying the software, as it makes the onboarding experience much smoother. Downloading a binary and not knowing what to do with it can be overwhelming for newbies. Advances users already know they have many different ways of doing it to fit their needs best and are less likely to just follow the official docs. 

So the review from me is that this looks very good, is concise, clear, very easy to follow and clearly highlights the few nuances that you should be aware of. Thumbs up 👍👍

I would like to add just a few nitpicks for your consideration.

1. I wouldn't call it `introduction`
Even with this tutorial in the docs, I think you should still read `Getting started ` first and naming it `introduction` makes it look like it's the first required read. So I would rename it to something like `Serving your first site tutorial` or `Simple Caddy production set up ` or similar. 

2. Add mention to `caddy fmt `and `validate`. 
On section `Your first site` after reloading the config with `sudo systemctl reload caddy` you mention what to do if the command fails, I think it will be great to mention that you can run `caddy fmt --overwrite /etc/caddy/Caddyfile` and `caddy validate --config /etc/caddy/Caddyfile` before triggering the reload to minimise silly mistakes.

3. Add quick mention to `caddy-dynamicdns`
On section `HTTPS` you say: 

> Before continuing, point your domain to your server's IP address. This means setting the value of your domain's A/AAAA record(s) to the public IP address of your server.

I would add quick note saying that caddy can do that for you using [caddy-dynamicdns](https://github.com/mholt/caddy-dynamicdns), not explaining how it's done, as it would add a bit of unnecessary noise to the tutorial but a quick mentions would help with discoverability of that very very handy plugin. 

4. Lastly, a very tiny one. Write at least once the full form of CA (certificate authority)
On section `Certificates` there are several mentions to CA, which maybe everyone should know what it means, but probably that's not the case and a quick googling of "CA" is not very helpful.

src/docs/markdown/introduction.md
```diff
@@ -0,0 +1,276 @@
...
+Oftentimes, your site consists of a backend application, but you want to put Caddy in front to handle TLS, routing, and other network-related details. Caddy's proxy is easy to use and extremely powerful.
+
+For example, if we have a backend that provides the site's API endpoints, we can easily proxy those with just 1 additional line of configuration:
+
+```caddy
+example.com {
+	root * /var/www/html
+	file_server
+	reverse_proxy /api/* 127.0.0.1:9000
+}
```
francislavoie, line 169: I think this would need a mention about directive sorting, otherwise it's not obvious why this would work (they might try using a matcher and assume it'll "just work" but it might not depending on the combination of directives).

src/docs/markdown/introduction.md
```diff
@@ -0,0 +1,276 @@
...
+
+Notice the `reverse_proxy` directive. The first argument, `/api/*`, is a [path matcher](/docs/caddyfile/matchers) which filters only requests within `/api/`. Then it proxies those to the backend app listening on :9000.
+
+If your backend is a PHP app, simply replace the `reverse_proxy` directive with the `php_fastcgi` directive:
+
+```caddy
+example.com {
+	root * /var/www/html
+	file_server
+	php_fastcgi /api/* 127.0.0.1:9000
```
francislavoie, line 180: I think this is misleading, because most of the time, this won't work.

For PHP apps, you need to configure `root` to your app's webroot, which should almost never be the actual root of the project, but actually a `public/` dir inside of the project. This is important for security so that random PHP scripts as part of the actual app can't be directly executed, or metadata/secret files (like a `.env`) can't be served by the file server.

And if you use a path matcher, it doesn't change the `try_files` logic within the directive, so it will still look for `index.php` at the configured `root`, and not inside of some `/api` directory within the project.

mholt:
Just FYI I'm considering this feedback for our new Caddy website which I'm working on.

PR #271, opened 2022-09-26
Fix basicauth documentation and example

clement-escolano:
Right now, the `basicauth` documentation says that the password hash for `basicauth` must be written as is. However, doing so results in an error when loading the config:

```
on: loading authentication providers: module name 'http_basic': provision http.authentication.providers.http_basic: base64-decoding password: illegal base64 data at input byte 0"}
```

base-64 encoding of the hash seems mandatory so this updates the documentation and the associated example


francislavoie:
What version of Caddy were you running? In v2.6.0 we just made a change to no longer require the redundant base64 encoding for bcrypt hashes. I think you just need to update Caddy, the docs are correct.

mholt:
Ah yeah, I'm pretty sure a Caddy upgrade is all that is needed. The docs reflect the latest version of Caddy.

clement-escolano:
Indeed it works fine now, sorry I had not updated before making the pull request.
Have a good day

mholt:
Excellent :) Thanks!

PR #272, opened 2022-09-30
Typo and word change in `root` documentation

elebow:


mholt (approved):
Thanks, that definitely reads better!

PR #274, opened 2022-10-03
docs: Various updates since 2.5.2 that got missed

francislavoie:
Did a big round of review looking at all the commits since v2.5.2 to catch anything we might have missed. Did a few other minor cleanups while I was at it.

FWIW, I didn't like the reordering of caddyfile placeholder shortcuts that was done earlier, I prefer them to be ordered by their expanded form instead of alphabetical by the shortcut. So I moved some of those around to feel more logical IMO.


mholt (commented):
Thanks! Overall looks good. I'm a bit confused by the reordering of the placeholders table though. The proposed ordering is not obvious, whereas alphabetical at least makes them easy to find, and still tends to group related placeholders.

francislavoie:
I think it is obvious. They're essentially hierarchical/alphabetical by their original placeholder. That way, they're together by context.

mholt:
I just don't think that's useful for most people who are looking for, "cookie" or "header" or "query" -- they have to scan instead of simply looking at the words in order...

I dunno. Should we take a survey?

francislavoie:
Think of the right side as categories. A cookie isn't a `uri`, it's easy to quickly scan past all those, etc.

mholt:
Ok, but maybe if we do that, we should switch the order of the columns. I don't think people (like me, anyway) scan the second column to find what they are looking for. Thoughts?

Jiehong:
My data point: I also read the first column first, and it _looks_ alphabetically sorted at first (up until "labels") and then you have slowdown it might not be.

Maybe it would make more sense to have the "Replaces" column first, ordered alphabetically and by category (but it needs to be called something else, like "Placeholder"), and "Shorthand" becomes the second column.

We _do_ read English from left to right, and top to bottom, so it feels natural.

francislavoie:
But left is "before", right is "after". I argue the same thing re English, left -> right. So I think the columns being in this current order makes the most lexical sense.

Every time I go looking for a placeholder, I do so looking by the module or area of the system first, drilling down the the exact thing I need. If they're alphabetical, then that becomes more difficult/tedious.

bbaa-bbaa:
Perhaps we should replace the allow_h2c here
https://github.com/caddyserver/website/blob/6370eb8c9cd256a8083db31d4342c53ae753bac7/src/docs/markdown/caddyfile/options.md?plain=1#L339-L341

mholt (approved):
Thanks Francis. Sorry about the confusion with the placeholder table. Let's give that a dedicated discussion in a new PR, because I think all 3 of us make valid points about how it should be displayed. Really appreciate this contribution!

PR #280, opened 2022-11-17
Fix reverseproxy http transport `dial_timeout` default value

francescopepe:
Changed the default value of the `dial_timeout` config of the httptransport to `3s` as it was introduced in v2.5.0 with this [PR](https://github.com/caddyserver/caddy/pull/4436/files#diff-3a7a275287ccf8af168eb64afaff75c8b96a1d231569609c35f2e5db488c1c8fL67).


francislavoie:
Oh, thanks! Not sure how we missed this :see_no_evil:

PR #284, opened 2022-12-10
docs: add a note on how to auto restart caddy on crashes with systemd

ncw:
It was decided in https://github.com/caddyserver/dist/pull/92 that the best way to auto restart caddy is using a systemd override file and this documents how to do that.


mholt (approved):
Awesome, thanks Nick! You beat me to it. LGTM

PR #286, opened 2022-12-21
chore: typo fix in v2

0xflotus:
Everywhere else in the docs you use the term `proxying`, only in v2.html you use `proxing`, so I fixed this to be consistent.


mholt:
Thanks, that was a typo

PR #288, opened 2023-01-06
reverse_proxy: Elaborate on upstream defaults, scheme defaults

francislavoie:
Some recent forum questions made me think this could be clarified a bit in this section to make it clear that Caddy proxies over HTTP by default. This can be confusing because Caddy as a server is HTTPS by default.


mholt (approved):

mholt (approved):
Thanks, these are good improvements and clarifications. Really appreciate it!

PR #290, opened 2023-01-13
docs: fix port number

protodeniz:
The privileged ports in Unix systems go from 0 to 1023 inclusive.

The doc here has an off-by-one error.  :)


mholt:
Good to know! I wasn't actually sure whether it was 1024 or 1023. I guess that makes sense. Cheers

PR #291, opened 2023-01-15
Add MySQL Config Adapter

zhangjiayin:


francislavoie:
Interesting!

Another approach you might want to consider taking, you could implement this as a config loader module which would allow this module to automatically reload the config when changes are made.

See https://caddyserver.com/docs/json/admin/config/load/

zhangjiayin:
Yes, I've checked the implementation: https://caddyserver.com/docs/json/admin/config/load/

The MySQL Config Adapter should reload the config automatically when changes are made according to the version value in the MySQL table, which has been implemented.

I think it should be more effective to implement the feature using an adapter than a config loader

the feature should be like this:
I'm trying to use the caddy server in containers, and add some domain host dynamics by inserting some value to the table
and updating the version in the table.
then the complete configuration should be reloaded automatically.
and l do not need to share any Caddyfile in Storage Class.

zhangjiayin:
and I've created a MySQL storage: https://pkg.go.dev/github.com/zhangjiayin/caddy-mysql-storage ( fork from https://github.com/yroc92/postgres-storage)
if you can add it to the doc too.

mholt:
So does this basically store the config in a MySQL table as a single JSON string? And when you change the config you are supposed to update the version number? or do you store multiple versions in the table and then you choose which one to use?

zhangjiayin:
it have serveral keys to be config should like below:

table

1. version:  if  configuration reload is needed, update the value (latest created value if used if multiple)
2. config:   the complete config of caddy server (latest created value if used if multiple)
3. config.admin: override the config.admin value of config if exists (latest created value if used if multiple)
4. config.storage: override the config.storage value of config if exists (latest created value if used if multiple)
5. config.logging: override the config.logging value of config if exists (latest created value if used if multiple)
6. config.apps: override the config.apps value of config if exists (latest created value if used if multiple)
7. config.apps.http.servers.srvXX.routes:  append the value  to config.apps.http.servers.srvXX.routes
your can dynamic add a host  by inserting a value with key  config.apps.http.servers.srvXX.routes to  MySQL  then update the version value

mholt (approved):
Awesome. Let's show it off, and hopefully it'll be useful for some!

PR #292, opened 2023-01-27
globalopts: Add `servers` name

danlsgiga:
Add the recently introduced `name` option to the global `servers` directive

https://github.com/caddyserver/caddy/issues/5322
https://github.com/caddyserver/caddy/pull/5323

### DIFF
--- a/src/docs/markdown/caddyfile/options.md
+++ b/src/docs/markdown/caddyfile/options.md
@@ -96,6 +96,7 @@ Possible options are:
 
 	# Server Options
 	servers [<listener_address>] {
+		name <name>
 		listener_wrappers {
 			<listener_wrappers...>
 		}
@@ -105,7 +106,6 @@ Possible options are:
 			write       <duration>
 			idle        <duration>
 		}
-		name <name>
 		metrics
 		max_header_size <size>
 		log_credentials
@@ -368,7 +368,7 @@ A custom name to assign to this server. Usually helpful to identify a server by 
 <aside class="tip">
 
 Keep in mind there's a caveat if you want to name your HTTP server and are using Auto-HTTPS. The server name config doesn't persist past adapting the config, and Auto-HTTPS happens at runtime, from the JSON config. 
-To overcome this, you'll need to create an empty `:80` or `http://` site block and rename it with `servers :80 { name http }` in the global options. With that Auto-HTTPS will add its redirect routes to that server.
+To overcome this, you'll need to create an empty `:80` or `http://` site block and set this option. With that Auto-HTTPS will add its redirect routes to that server.
 
 </aside>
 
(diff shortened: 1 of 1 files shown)


francislavoie:
Thanks! We'll need an `<aside class="tip">` here which mentions the HTTP server caveat I mentioned in the issue.

src/docs/markdown/caddyfile/options.md
```diff
@@ -105,6 +105,7 @@ Possible options are:
 			write       <duration>
 			idle        <duration>
 		}
+		name main
```
```suggestion
		name <name>
```

francislavoie: Also I think we should move this to the top of `servers`.

src/docs/markdown/caddyfile/options.md
```diff
@@ -360,6 +361,30 @@ For example, to configure different options for the servers on ports `:80` and `
...
 ```
 
+##### `name`
+
+A custom name to assign to this server. Usually helpful to identify a server by its name in logs and metrics. If not set, Caddy will define it dynamically using a `srvX` pattern, where `X` starts with 0 and increments based on the number of servers in the config.
+
+<aside class="tip">
+
+Keep in mind there's a caveat if you want to name your HTTP server and are using Auto-HTTPS. The server name config doesn't persist past adapting the config, and Auto-HTTPS happens at runtime, from the JSON config. 
+To overcome this, you'll need to create an empty `:80` or `http://` site block and rename it with `servers :80 { name http }` in the global options. With that Auto-HTTPS will add its redirect routes to that server.
```
francislavoie: Having `servers :80 { name http }` in the docs is misleading because that's not valid config; blocks must have newlines. I might have said it colloquially in the other issue, but it was just to make it a natural sentence :wink:

src/docs/markdown/caddyfile/options.md
```diff
@@ -105,6 +105,7 @@ Possible options are:
 			write       <duration>
 			idle        <duration>
 		}
+		name <name>
```
francislavoie: Should move this up as well

src/docs/markdown/caddyfile/options.md
```diff
@@ -360,6 +361,30 @@ For example, to configure different options for the servers on ports `:80` and `
...
+
+```caddy
+{
+	servers :80 {
+		name http-auto-redirect
+	}
+}
+
+:80
+```
```
francislavoie: We can flesh out the example a bit more

```suggestion
```caddy
{
	servers :443 {
		name https
	}
	servers :80 {
		name http
	}
}

example.com {
}

http:// {
}
```
```

francislavoie: oh that's dumb... github suggestions don't work properly with code blocks because they use backticks for those as well :man_facepalming: don't click commit on this one, it's missing the ``` at the end

danlsgiga: I got you! New commit incoming shortly!

Issue #293, opened 2023-01-27
Document internal directive order


mpldr:
The internal directive ordering is – I assume – falling on my toes right now. Would it be possible to add the order to the Docs?

mholt:
You mean this order? https://caddyserver.com/docs/caddyfile/directives#directive-order

mpldr:
Yes indeed. Look like I've been blind. Thank you very much :)

PR #294, opened 2023-01-27
Add additional references to the directive order

mpldr:
Depending on which directives you look at, there are no links to the directive order. This adds a reference to some of the most commonly used directive-docs.

References: #293


francislavoie (approved):
Thanks! Those are some good spots to make those callouts :+1:

PR #297, opened 2023-02-09
options: Add note about server listener_address matching

danlsgiga:
This is something I just figured out when testing the new `name` option in the `servers` block.

Just adds a tip explaining that an exact match is required to apply settings to specific listeners when using `bind` and `default_bind` directives

### DIFF
--- a/src/docs/markdown/caddyfile/options.md
+++ b/src/docs/markdown/caddyfile/options.md
@@ -410,11 +410,13 @@ For example:
 
 ```caddy
 {
-	servers :8080 {  <--- THIS WILL NOT MATCH ANY LISTENER
+	# This will NOT match any listener
+	servers :8080 {
 		name private
 	}
 	
-	servers 192.168.1.2:8080 {  <--- THIS WILL WORK BECAUSE IS AN EXACT MATCH 
+	# This will work because it's an exact match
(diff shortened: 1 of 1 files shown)


src/docs/markdown/caddyfile/options.md
```diff
@@ -400,6 +400,33 @@ http:// {
...
+	
+For example:
+
+```caddy
+{
+	servers :8080 {  <--- THIS WILL NOT MATCH ANY LISTENER
+		name private
+	}
+	
+	servers 192.168.1.2:8080 {  <--- THIS WILL WORK BECAUSE IS AN EXACT MATCH
```
francislavoie: Let's use Caddyfile comments so that it gets highlighted correctly. Also we don't need to use caps, it's too aggressive to read in docs.

```suggestion
	# This will NOT match any listener
	servers :8080 {
		name private
	}
	
	# This will work because it's an exact match
	servers 192.168.1.2:8080 { 
```

danlsgiga: Thanks, makes total sense!

francislavoie:
I reworked it a bit, I moved the new aside to earlier, reworded it a bit, and put the example inside the `<aside>`. I wasn't sure that would look good, but I think it works!

Thanks, merging

PR #302, opened 2023-02-19
Update rewrite.md

769344359:
it will be error if use the example
error parsing regexp: missing argument to repetition operator: `*`


francislavoie:
Not true. The example is correct.

There's no regexp involved here. The `*` is a request matcher token: https://caddyserver.com/docs/caddyfile/matchers#syntax

PR #309, opened 2023-03-18
Fix link to global trusted proxies page

JackBailey:


francislavoie:
Oh, thanks! Good catch. Markdown is funny sometimes with the anchor tags it generates :see_no_evil:

PR #310, opened 2023-03-20
Introduce tabbed sections in Markdown, Docker docs

francislavoie:
For quite a while now, I've wanted a way to have sections in the docs that are platform-specific hidden behind tabs.

This is super easy to do with a library like https://alpinejs.dev since it allows us to do it entirely inline within markup.

Preview of what it looks like (GIF):

firefox_IZ2mblLKF1

- Pulling Alpine from CDNs because it's lower maintenance to do so and we get the latest patch versions automatically which is nice.
- We use `x-data` to define the default value for the tab. Then, `x-on:click` is used to update the value, and `x-show` is a condition to only show the tab content if the chosen tab matches. `x-bind:class` is used to add the `active` class for the current tab.
- For the tab's data, since we care to show the user the docs for their current OS, I'm reusing the platform detection logic we had for `download.html` but I'm moving it to `common.js` so that I can use it here as well. Made a helper `defaultOS()` since the detected OS might be something else like `solaris` (lol) which we don't have instructions for, so I need to validate against the ones supported by the tabs.
- I pulled in the `persist` plugin as well https://alpinejs.dev/plugins/persist which adds the `$persist()` function to make data persist across page navigation. When we expand the usage of this, it would mean that we could share the data across all pages etc.
- Notice that I put newlines around the content of the `<div x-show>`. This is so that the content within is rendered as Markdown (obviously in this case there's no "markdown" since it's just a `<pre><code>`). In general, there would be textual content and we want that to be caught by our markdown parser so those newlines are key to making that work.
- I took a shot at doing some CSS that makes sense and fits with the docs theme. The colors and border width could use some tweaking maybe. I hand picked the dark theme color to be lightly darker than the border we use for `<h1>` because that was too bright for this. The border color for light theme is pretty uninspired, I'm not sure what to do with it; should it be blue tinted as well? I use the dark theme so I have a better feel for how the dark theme should look. Feel free to fix the CSS on this branch @mholt

So that's the spiel about Alpine. But in addition to that, this PR's triggering motivation was a want to improve the Docker docs a bit:

- I found https://github.com/dunglas/symfony-docker/blob/main/docs/tls.md earlier today, they mention some commands that can be used to install the root CA cert on the host machine. These are nice and clean commands, so I wanted to pull that idea into our docs because it's a question we get a lot. /cc @dunglas thanks for that (I don't think you authored it but it is your repo 😅)

  I've always struggled with how to make this easier for users. My thought for the past couple years was "we have `caddy trust`, how can we make it possible to users to run that on their host machine?", but the reality is, that's not really viable. It would mean they'd need to have a Caddy binary on the host as well, but they're using Docker specifically so they don't need it on the host. So this is essentially me giving up on that idea. I've been telling people "google it" for how to install the cert on their platform when asked on the forums, but that's not great.

  The other limitation here is obviously that this doesn't attempt to install the root CA cert in browser trust stores. We might need to augment the documented command to do that at some point. I dunno. Seems annoying 😬. It's essentially redoing work that https://github.com/smallstep/truststore already handles. Bah.

- Docker Compose V2 doesn't use a `-` anymore, wanted to clarify that clearly.

- Some beginners might not have a clue what the docker-compose.yml does, figured I could add a bit of explanation about each part.

- Added an aside noting that `localhost` doesn't do what many users expect in Docker for `reverse_proxy`.

- Added a bit more to the logs command to make it more useful by default. Typically users want their most recent logs, not their oldest. So using `-n` and `-f` helps with that.

### DIFF
--- a/src/docs/markdown/running.md
+++ b/src/docs/markdown/running.md
@@ -14,6 +14,8 @@ While Caddy can be run directly with its [command line interface](/docs/command-
   - [Overrides](#overrides)
 - [Windows Service](#windows-service)
 - [Docker Compose](#docker-compose)
+  - [Setup](#setup)
+  - [Usage](#usage)
 
 
 ## Linux Service
@@ -105,7 +107,9 @@ You can stop the service with:
 <pre><code class="cmd bash">sudo systemctl stop caddy</code></pre>
 
 <aside class="advice">
-	Do not stop the service to change Caddy's configuration. Stopping the server will incur downtime. Use the reload command instead.
+
+Do not stop the service to change Caddy's configuration. Stopping the server will incur downtime. Use the reload command instead.
+
 </aside>
 
 The Caddy process will run as the `caddy` user, which has its `$HOME` set to `/var/lib/caddy`. This means that:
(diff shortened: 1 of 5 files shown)


mholt (approved):
Wow, this is a really well-thought-out change, Francis. Thank you for the improvements!

A hearty **accept** from me. I think it looks fine as-is, personally. And I'll be updating the site with a new design this year, so I'll bring these tabs into the new design (but with different styles most likely, to match).

(PS. Am I the only one who finds it annoying that you can't approve from the conversation view too)

francislavoie:
Great! 😊

> (PS. Am I the only one who finds it annoying that you can't approve from the conversation view too)

Yeah I agree. But it does kinda help funnel users to actually look at the code first instead of blindly approving things. It's kinda a speedbump, probably an intentional one. We're driving in a school zone 🤣

PR #315, opened 2023-04-14
handle_errors: use `replace_status` for proxy example

lemniskett:
Hi, I actually used the `http.cat` handle_error snippet in my website and realized that any errors are logged as 200 in access logs. adding `replace_status` directive fixed the issue.


francislavoie:
Nice, yeah good point. We should probably go a step further and document the same thing for `file_server` and `respond` which have similar ways of overriding the status code.

Edit: Nevermind that's not necessary, both `respond` and `file_server` already inherit the error's status code without needing an override.

PR #317, opened 2023-04-29
Add clarification to `acme_ca` about `.local` TLDs

mrexodia:
Discussion in https://github.com/caddyserver/caddy/issues/5512

### DIFF
--- a/src/docs/markdown/caddyfile/options.md
+++ b/src/docs/markdown/caddyfile/options.md
@@ -295,7 +295,7 @@ Skips the attempts to install the local CA's root into the system trust store, a
 ##### `acme_ca`
 Specifies the URL to the ACME CA's directory. It is strongly recommended to set this to Let's Encrypt's [staging endpoint](https://letsencrypt.org/docs/staging-environment/) for testing or development. Default: ZeroSSL and Let's Encrypt's production endpoints.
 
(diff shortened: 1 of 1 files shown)


mholt (commented):
I wonder if instead of repeating this information, we should just link to it and make the user aware to check that. Because now I'm considering changing the behavior for `.local`.

mrexodia:
This sentence would have saved me quite a bit of time. I agree it’s duplication, but if you decide to change the behavior it would be a small effort to update it.

src/docs/markdown/caddyfile/options.md
```diff
@@ -295,6 +295,7 @@ Skips the attempts to install the local CA's root into the system trust store, a
 ##### `acme_ca`
 Specifies the URL to the ACME CA's directory. It is strongly recommended to set this to Let's Encrypt's [staging endpoint](https://letsencrypt.org/docs/staging-environment/) for testing or development. Default: ZeroSSL and Let's Encrypt's production endpoints.
 
+Note that the [hostname requirements](/docs/automatic-https#hostname-requirements) apply, and ACME CA directory will not be used for IP addresses, `localhost` and hostnames with a `.local` TLD.
```
```suggestion
Note that a globally-configured ACME CA may not apply to all sites; see the [hostname requirements](/docs/automatic-https#hostname-requirements) for using the default ACME issuer(s).
```

What about something like this? The sentence can remain but we link to the relevant docs instead of duplicating them.

mrexodia: Yeah I’m good with anything, it’s your project ^^

mholt (approved):
Thanks for the contribution!

PR #321, opened 2023-05-11
reverse_proxy: rewrite method to GET for X-Accel-Redirect support.

lemniskett:
There are cases where non-GET-or-HEAD requests will need to send files using X-Accel-Redirect, which will caddy reject with `405 Method not allowed` since file_server only accepts GET and HEAD request (https://github.com/caddyserver/caddy/commit/a3ae146cbdf2cfbdbdf5feea52e8bf407cce2b31)

This is based on https://github.com/caddyserver/caddy/issues/5529


mholt (approved):
LGTM -- thanks for the enhancement!

PR #323, opened 2023-05-17
Add information for how to add a new module to the module documentation

znkr:
Something like this would have helped me find the account page. I might have missed the documentation somewhere else though. TBH, I am not sure if this is a good fit where I put it. Happy to change direction or abandon this PR.


mholt (approved):
Thanks!

FYI I'm working on a new Caddy website and mostly-all-new-docs that will cover this better, too :)

PR #325, opened 2023-06-05
Change v1 Docs icon from external link to download

optimorg:
I clicked this accidentally believing I'd be viewing another page


francislavoie:
Please revert the other whitespace changes, they aren't relevant

mholt:
Thanks for the contribution. I'm happy to accepting this, but just FYI I've started work on a new site where this change won't be relevant -- so it will be short-lived. Is that OK?

PR #333, opened 2023-07-15
cli: document the new command `file-server export-template`

mohammed90:
I'm not sure if there's a better way to document sub-command. Any other thoughts beside plainly listing it like any other?

### DIFF
--- a/src/docs/markdown/caddyfile/directives/file_server.md
+++ b/src/docs/markdown/caddyfile/directives/file_server.md
@@ -48,7 +48,7 @@ file_server [<matcher>] [browse] {
 
 - **browse** <span id="browse"/> enables file listings for requests to directories that do not have an index file.
 
-  - **<template_file>** <span id="template_file"/> is an optional custom template file to use for directory listings. Defaults to the template that can be found [here in the source code ![external link](/resources/images/external-link.svg)](https://github.com/caddyserver/caddy/blob/master/modules/caddyhttp/fileserver/browse.html). Browse templates can use actions from [the standard templates module](/docs/modules/http.handlers.templates#docs) as well.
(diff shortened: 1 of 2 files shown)


francislavoie:
Yeah I think doing it like a normal top-level command is fine.

I think we should also document it on the file_server directive page, for visibility.

src/docs/markdown/caddyfile/directives/file_server.md
```diff
@@ -48,7 +48,7 @@ file_server [<matcher>] [browse] {
 
 - **browse** <span id="browse"/> enables file listings for requests to directories that do not have an index file.
 
-  - **<template_file>** <span id="template_file"/> is an optional custom template file to use for directory listings. Defaults to the template that can be found [here in the source code ![external link](/resources/images/external-link.svg)](https://github.com/caddyserver/caddy/blob/master/modules/caddyhttp/fileserver/browse.html). Browse templates can use actions from [the standard templates module](/docs/modules/http.handlers.templates#docs) as well.
+  - **<template_file>** <span id="template_file"/> is an optional custom template file to use for directory listings. Defaults to the template that can be extracted using the command `caddy file-server export-template`, which will print the defaut template to stdout. Browse templates can use actions from [the standard templates module](/docs/modules/http.handlers.templates#docs) as well.
```
mohammed90: > I think we should also document it on the file_server directive page, for visibility.

@francislavoie, I mentioned it here. Should I also add an example at the bottom?

francislavoie: Oh sorry, missed that. Mobile things.

francislavoie: Actually let's keep both the original link and the command. Having two options to get it is fine. The command might not be convenient to use for some users (like when Caddy isn't directly accessible and is under somekind of automation)

mohammed90: Done.

mholt (approved):
Oh, thanks for this -- yeah I think this'll work well. It's our first "sub sub command" so it's probably fine :+1: Thank you!

francislavoie (approved):

PR #335, opened 2023-08-06
running: Notes about SELinux

lldb3:
Hello,

I've added minor documentation as to how you would install a systemd Unit file for Caddy on an SELinux enabled system.

This was tested on Fedora, I unfortunately don't have another distro to try it on.

In any case, I hope this helps, and thanks for all the existing doc' !

### DIFF
--- a/src/docs/markdown/running.md
+++ b/src/docs/markdown/running.md
@@ -7,19 +7,18 @@ title: Keep Caddy Running
 While Caddy can be run directly with its [command line interface](/docs/command-line), there are numerous advantages to using a service manager to keep it running, such as ensuring it starts automatically when the system reboots and to capture stdout/stderr logs.
 
 
-- [Keep Caddy Running](#keep-caddy-running)
-  - [Linux Service](#linux-service)
-    - [Unit Files](#unit-files)
-    - [Manual Installation](#manual-installation)
-    - [Using the Service](#using-the-service)
-    - [Overrides](#overrides)
-    - [SELinux System Considerations](#selinux-system-considerations)
-  - [Windows service](#windows-service)
-    - [sc.exe](#scexe)
-    - [WinSW](#winsw)
-  - [Docker Compose](#docker-compose)
-    - [Setup](#setup)
-    - [Usage](#usage)
+- [Linux Service](#linux-service)
+  - [Unit Files](#unit-files)
+  - [Manual Installation](#manual-installation)
+  - [Using the Service](#using-the-service)
+  - [Overrides](#overrides)
+  - [SELinux System Considerations](#selinux-system-considerations)
+- [Windows service](#windows-service)
+  - [sc.exe](#scexe)
+  - [WinSW](#winsw)
+- [Docker Compose](#docker-compose)
+  - [Setup](#setup)
+  - [Usage](#usage)
 
 
 ## Linux Service
@@ -156,7 +155,7 @@ RestartSec=5s
 Then, save the file and exit the text editor, and restart the service for it to take effect:
 <pre><code class="cmd bash">sudo systemctl restart caddy</code></pre>
 
-### SELinux System Considerations
+### SELinux Considerations
 
 On SELinux enabled systems you have two options:
 1. Install Caddy using the [COPR repo](https://copr.fedorainfracloud.org/coprs/g/caddy/caddy/). Your systemd file and caddy binary will be created and labelled correctly. If you wish to use a custom build of Caddy, you'll need to label the executable as described below.
@@ -167,8 +166,7 @@ Systemd unit files and their executables will not be run unless labelled with `s
 The `systemd_unit_file_t` is automatically applied to files created in `/etc/systemd/...`, so be sure to create your `caddy.service` file there.
 
 To tag the caddy binary, you can use the following commands:
-<pre><code class="cmd bash">semanage fcontext -a -t bin_t /usr/bin/caddy
-restorecon -Rv /usr/bin/caddy
+<pre><code class="cmd bash">semanage fcontext -a -t bin_t /usr/bin/caddy && restorecon -Rv /usr/bin/caddy
 </code></pre>
 
 ## Windows service
--- a/src/docs/markdown/running.md
+++ b/src/docs/markdown/running.md
@@ -12,6 +12,7 @@ While Caddy can be run directly with its [command line interface](/docs/command-
   - [Manual Installation](#manual-installation)
   - [Using the Service](#using-the-service)
   - [Overrides](#overrides)
+  - [SELinux System Considerations](#selinux-system-considerations)
 - [Windows Service](#windows-service)
 - [Docker Compose](#docker-compose)
   - [Setup](#setup)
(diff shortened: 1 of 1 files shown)


francislavoie:
Thanks.

Any reason you aren't using the COPR repo on Fedora?

lldb3:
Honestly I didn't try it, and went straight to using the xcaddy command to build it with custom modules.

The SELinux issue is probably solved using COPR directly, but then you'd have to update the binary with your own modules and face the same issue (custom binary improperly labelled in SELinux).

**Update**: I just added a new commit with shorter instructions, the previous ones worked but you could not enable the service (because symlinked).

src/docs/markdown/running.md
```diff
@@ -152,6 +153,18 @@ RestartSec=5s
 Then, save the file and exit the text editor, and restart the service for it to take effect:
 <pre><code class="cmd bash">sudo systemctl restart caddy</code></pre>
 
+### SELinux System Considerations
+
+On SELinux enabled systems, systemd unit files and their executables will not be run unless labelled with `systemd_unit_file_t` and `bin_t` respectively.
```
mholt: Can/should we add a note that if installing Caddy from the COPRs, this is taken care of for you?

mholt (approved):
I don't have any SELinux experience personally but it looks like this could be helpful. I do wonder if we should let people know that it's only needed if installing Caddy manually.

lldb3:
Sorry for the late reply.

@mholt I've merged with your latest updates and added a note about COPR, hope it works.
Cheers, Elmø

src/docs/markdown/running.md
```diff
@@ -152,6 +156,23 @@ RestartSec=5s
 Then, save the file and exit the text editor, and restart the service for it to take effect:
 <pre><code class="cmd bash">sudo systemctl restart caddy</code></pre>
 
+### SELinux System Considerations
+
+On SELinux enabled systems you have two options:
+1. Compile Caddy and label the files yourself
+2. Install Caddy using the [COPR repo](https://copr.fedorainfracloud.org/coprs/g/caddy/caddy/), your systemd file and caddy binary will be created and labelled correctly. NB: if you want to recompile Caddy with modules, you will have to label the executable as described bellow.
```
francislavoie: I suggest reordering these lines so we're recommending installing from COPR first.

Also:

- `NB:` is not necessary, just start the sentence with `If`
- `bellow` -> `below`.
- I'd write it "If you wish to use a custom build of Caddy, you'll need to..."

Also users may not always have compiled Caddy themselves, they might have used the `/download` page or grabbed a vanilla build from GitHub so we might want to use different wording to cover that too.

src/docs/markdown/running.md
```diff
@@ -152,6 +156,23 @@ RestartSec=5s
...
+   
+
+Systemd unit files and their executables will not be run unless labelled with `systemd_unit_file_t` and `bin_t` respectively.
+
+The `systemd_unit_file_t` is automatically applied to files created in `/etc/systemd/...`, so be sure to create your `caddy.service` file there.
+
+To tag the caddy binary, you can use the following commands:
+```shell
```
francislavoie: Please take a look at how we're doing code blocks for commands, i.e. HTML so that it has the CSS classes we want to have, `cmd bash`

lldb3: Thanks, will push in the next commit !

src/docs/markdown/running.md
lldb3: I'll test the `/download` case tomorrow on a fedora machine and see how it behaves, since I didn't package it I can't speculate on how it works yet.

Thank you for the typo correction, I've updated them and will push tomorrow after I test the direct download case

francislavoie: All the download page gives you is a Caddy binary with whatever plugins you choose. It's the same output as using xcaddy, in fact it's just a wrapper around xcaddy.

src/docs/markdown/running.md
```diff
@@ -152,6 +156,20 @@ RestartSec=5s
 Then, save the file and exit the text editor, and restart the service for it to take effect:
 <pre><code class="cmd bash">sudo systemctl restart caddy</code></pre>
 
+### SELinux System Considerations
+
+On SELinux enabled systems you have two options:
+1. Install Caddy using the [COPR repo](https://copr.fedorainfracloud.org/coprs/g/caddy/caddy/), your systemd file and caddy binary will be created and labelled correctly. If you wish to use a custom build of Caddy, you'll need to label the executable as described below.
```
mholt:
```suggestion
1. Install Caddy using the [COPR repo](https://copr.fedorainfracloud.org/coprs/g/caddy/caddy/). Your systemd file and caddy binary will be created and labelled correctly. If you wish to use a custom build of Caddy, you'll need to label the executable as described below.
```

Had a hard time parsing this sentence at first :sweat_smile:

src/docs/markdown/running.md
```diff
@@ -152,6 +156,20 @@ RestartSec=5s
...
+2. [Download](https://caddyserver.com/download) Caddy from the site, or compile it with *xcaddy*. In both cases you will need to label the files yourself.
```
mholt:
```suggestion
2. [Download Caddy from this site](https://caddyserver.com/download) or compile it with [`xcaddy`](https://github.com/caddyserver/xcaddy). In both cases you will need to label the files yourself.
```

mholt (changes requested):
Just some minor nits, otherwise LGTM and we can merge if you agree with these slight changes :+1:

src/docs/markdown/running.md
```diff
@@ -7,15 +7,19 @@ title: Keep Caddy Running
...
-- [Linux Service](#linux-service)
-  - [Unit Files](#unit-files)
-  - [Manual Installation](#manual-installation)
-  - [Using the Service](#using-the-service)
-  - [Overrides](#overrides)
-- [Windows Service](#windows-service)
-- [Docker Compose](#docker-compose)
-  - [Setup](#setup)
-  - [Usage](#usage)
+- [Keep Caddy Running](#keep-caddy-running)
```
francislavoie: I don't think we need this extra level in the TOC, it's kinda just noise.

src/docs/markdown/running.md
```diff
@@ -7,15 +7,19 @@ title: Keep Caddy Running
...
+  - [Linux Service](#linux-service)
+    - [Unit Files](#unit-files)
+    - [Manual Installation](#manual-installation)
+    - [Using the Service](#using-the-service)
+    - [Overrides](#overrides)
+    - [SELinux System Considerations](#selinux-system-considerations)
```
francislavoie: Maybe just `SELinux Considerations`? The word `System` feels redundant.

src/docs/markdown/running.md
```diff
@@ -152,6 +156,20 @@ RestartSec=5s
...
+2. [Download Caddy from this site](https://caddyserver.com/download) or compile it with [`xcaddy`](https://github.com/caddyserver/xcaddy). In both cases you will need to label the files yourself.
+
+Systemd unit files and their executables will not be run unless labelled with `systemd_unit_file_t` and `bin_t` respectively.
+
+The `systemd_unit_file_t` is automatically applied to files created in `/etc/systemd/...`, so be sure to create your `caddy.service` file there.
+
+To tag the caddy binary, you can use the following commands:
+<pre><code class="cmd bash">semanage fcontext -a -t bin_t /usr/bin/caddy
+restorecon -Rv /usr/bin/caddy
+</code></pre>
```
francislavoie: I think making it a one-liner is helpful here:
```suggestion
<pre><code class="cmd bash">semanage fcontext -a -t bin_t /usr/bin/caddy && restorecon -Rv /usr/bin/caddy
</code></pre>
```

I think it would render incorrectly on two lines (it would only show `$` on the first line with the CSS) unless we used a span for each line, like the below (if you don't like `&&`):

```suggestion
<pre><code class="cmd"><span class="bash">semanage fcontext -a -t bin_t /usr/bin/caddy</span>
<span class="bash">restorecon -Rv /usr/bin/caddy</span</code></pre>
```

lldb3: A oneliner is fine as long as it works ^^

src/docs/markdown/running.md
lldb3: Arf, my bad it's vscode's Markdown All in One that automatically generates a Table of content on save..

lldb3:
Okay its updated! Good catch on the TOC content VSCode added, it just slipped by unnoticed by me.

mholt (approved):
Thanks! LGTM now.

francislavoie (approved):
LGTM as well

mholt:
Thank you for the contribution!

francislavoie:
https://caddyserver.com/docs/running#selinux-considerations

lldb3:
Happy to, hope it helps some lost souls out there :D

PR #336, opened 2023-08-10
Better styles and ux

AmagiDDmxh:
1. Add .current class to the current link
2. sticky sidebar

### DIFF
--- a/src/includes/docs/head.html
+++ b/src/includes/docs/head.html
@@ -8,3 +8,15 @@
 <script type="text/javascript">window.CaddyfileDirectives = {{$directives | toJson}};</script>
 <script src="/resources/js/jquery-3.4.1.min.js"></script>
 <script src="/resources/js/docs.js"></script>
+<script type="text/javascript">
+	$(function(){
+		var current = location.pathname;
+		$('header > nav a').each(function(){
+			var $this = $(this);
+			// if the current path is like this link, make it active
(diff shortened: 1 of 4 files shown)


francislavoie:
> Add .current class to the current link

It already does this. I don't understand what you're trying to fix.

> sticky sidebar

I'm not sure I like this, because it adds a scroll bar to the sidebar.

Either way, @mholt is working on a redesign, so these changes will effectively be thrown away soon once that's completed.

mholt:
Yeah; I agree, I'm not quite sure what the problem is.

AmagiDDmxh:
> It already does this. I don't understand what you're trying to fix.

There were two implementations of **.current class active** in `src/resources/js/docs.js` and in `src/resources/js/account/common.js`. However, none of them are working as expected.

The sticky sidebar can be deleted. 

Thank you very much for mentioning the redesign. So this could be a closed pr.

francislavoie:
> However, none of them are working as expected.

I don't agree? It works fine. Notice `Directives` is highlighted as the current page.

AmagiDDmxh:
> > However, none of them are working as expected.
> 
> I don't agree? It works fine. Notice `Directives` is highlighted as the current page.
> 
> 

But it didn't work on header-nav

Before at /docs page

CleanShot 2023-08-12 at 17 47 22

After

CleanShot 2023-08-12 at 17 48 03

PR #337, opened 2023-08-10
Implement cache busting

francislavoie:
- Switch all `head.html` to be rendered with `import` instead of `include`, to retain the `define` template stack

- Define a `cacheBust` template in the main `head.html` whose value is pulled from `src/.commit-hash` if that file exists. The file gets created on the website's server with a `post-checkout` git-hook. Its value is something like `?v=abc1234`, i.e. a query. If the file isn't there (e.g. in local dev), the template is empty string (no query).

- Use `{{template "cacheBust"}}` as a suffix to any JS/CSS asset files which are likely to change. I skipped minified ones cause they mostly had version numbers in the filename so it's redundant (if we upgrade those libs, we'll change the filename anyway).

I'm not sure if there's a better way. This is what I could get to work. I'm only dabbling in templates-land, I'm not too familiar with what's conventional. Is this fine? A `cacheBust` template is the only way I could figure out how to load the file once per request, because variables like `$cacheBust := "foo"` are not propagated across template files. 

To test this, checkout this branch, launch the website, and look at the raw HTML output to see that files don't have `?v=`. Then, create the `src/.commit-hash` file, put in `some-dumb-string`. Then refresh the website, you'll see `?v=some-dumb-string` appended to most JS/CSS files.

### DIFF
--- a/src/account/create.html
+++ b/src/account/create.html
@@ -2,8 +2,9 @@
 <html>
 	<head>
 		<title>Create Account - Caddy</title>
-		{{include "/includes/account/head.html"}}
-		<script src="/resources/js/account/create.js"></script>
+		{{import "/includes/account/head.html"}}
+		{{template "account-head"}}
+		<script src="/resources/js/account/create.js{{template "cacheBust"}}"></script>
 	</head>
 	<body>
 		<form action="/api/create-account" class="card">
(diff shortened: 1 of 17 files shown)


mholt (approved):
This is very clever! :smiley:

francislavoie:
Deployed! It works:

image

mholt:
YAYYY :raised_hands:

PR #338, opened 2023-08-11
Add copy to clipboard feature

danditomaso:
This adds the copy to clipboard feature that is available on many documentation websites. It sits inline with the text, so there wont be any cases where it will overlap existing content. 

Clicking the copy button takes the contents inside the code block and copies it to your clipboard, you receive a tooltip when clicking it that disappears after a second. I'm open to tweaking the timing on the tooltip if it feels slow.

### DIFF
--- a/src/resources/css/docs.css
+++ b/src/resources/css/docs.css
@@ -80,7 +80,7 @@ header nav .button {
 	text-decoration: none;
 }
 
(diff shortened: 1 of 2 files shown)


danditomaso:
Closing branch because it was based on `new` instead of `master`

PR #339, opened 2023-08-11
add copy to clipboard functionality

danditomaso:
This adds the copy to clipboard feature that is available on many documentation websites. It sits inline with the text, so there wont be any cases where it will overlap existing content.

Clicking the copy button takes the contents inside the code block and copies it to your clipboard, you receive a tooltip when clicking it that disappears after a second. I'm open to tweaking the timing on the tooltip if it feels slow.

### DIFF
--- a/src/resources/css/docs.css
+++ b/src/resources/css/docs.css
@@ -80,7 +80,7 @@ header nav .button {
 	text-decoration: none;
 }
 
-.breadcrumb-siblings a+a {
+.breadcrumb-siblings a + a {
 	padding-top: 10px;
 }
 
@@ -113,7 +113,7 @@ main > nav.sidebar {
 main nav li img,
 article li img,
 img.external-link {
-	max-height: .9em;
+	max-height: 0.9em;
 }
 
 main nav ul {
@@ -133,26 +133,24 @@ main nav li a {
 	color: #546c75;
 }
 
-
-
 main nav li a:hover {
 	color: #01324b;
 }
 
 main nav li a:before {
-	content: '\203A';
+	content: "\203A";
 	font-weight: bold;
 	font-size: 20px;
 	line-height: 1rem;
 	position: absolute;
 	opacity: 0;
 	left: 0;
-	transition: left .15s, opacity .15s;
+	transition: left 0.15s, opacity 0.15s;
 }
 
 main nav li a:hover:before {
 	opacity: 1;
-	left: .75em;
+	left: 0.75em;
 }
 
 main nav li li a:hover:before {
@@ -187,7 +185,7 @@ main nav li li a {
 
 	background: white;
 	border-radius: 4px;
-	box-shadow: 1px 2px 4px 1px rgba(0, 0, 0, .15);
+	box-shadow: 1px 2px 4px 1px rgba(0, 0, 0, 0.15);
 }
 
 #paper1,
@@ -258,8 +256,12 @@ article pre,
 article table {
 	margin-bottom: 1.5rem;
 }
-article > .fullwidth { margin-bottom: 1.5rem; }
-article > .fullwidth > * { margin-bottom: 0; }
+article > .fullwidth {
+	margin-bottom: 1.5rem;
+}
+article > .fullwidth > * {
+	margin-bottom: 0;
+}
 
 article > pre.chroma > code {
 	background: none;
@@ -292,7 +294,7 @@ article li {
 article li p,
 article li ul,
 article li ol {
-	margin-bottom: .5em;
+	margin-bottom: 0.5em;
 }
 
 h1,
@@ -341,10 +343,10 @@ h5 {
 
 .anchor-link {
 	opacity: 0;
-	font-size: .6em;
+	font-size: 0.6em;
 	border-radius: 10px;
-	padding: .3em .5em;
-	margin-left: .25em;
+	padding: 0.3em 0.5em;
+	margin-left: 0.25em;
 	position: absolute;
 	top: 5px;
 }
@@ -356,7 +358,7 @@ h5 {
 }
 
 .anchor-link:hover {
-	background-color: rgba(0, 0, 0, .075);
+	background-color: rgba(0, 0, 0, 0.075);
 }
 
 code {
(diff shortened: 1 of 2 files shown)


francislavoie:
Hmm, this had some side-effects on `/docs/install`. For example, the Debian section's commands have a lot more padding now.

Before:

image

After:

image

Also I'm only seeing the copy button on some of the blocks on this page. I'm not sure why though, they aren't significantly different other than missing the button :thinking: Probably just a selector bug.

src/resources/css/docs.css
```diff
@@ -133,26 +133,24 @@ main nav li a {
 	color: #546c75;
 }
 
-
-
```
francislavoie, line 137: These extra empty lines were kinda intentional to split apart the file in sections. We should probably add comments between these sections instead I figure.

src/resources/js/docs.js
```diff
@@ -15,11 +17,17 @@ $(function() {
...
 	}
-	$currentPageLink.addClass('current');
+	$currentPageLink.addClass("current");
 
 	// add anchor links, inspired by https://github.com/bryanbraun/anchorjs
-	$('article > h1[id], article > h2[id], article > h3[id], article > h4[id], article > h5[id], article > h6[id]').each(function() {
-		var $anchor = $('<a href="#'+this.id+'" class="anchor-link" title="Direct link">🔗</a>');
+	$(
+		"article > h1[id], article > h2[id], article > h3[id], article > h4[id], article > h5[id], article > h6[id]"
+	).each(function () {
```
francislavoie, line 25: I assume you used Prettier? It does some truly weird stuff with line length. Not a fan. :man_shrugging: 

In general I appreciate the consistency cleanup (thanks for that), just the line length stuff I find frustrating. To me it does exactly the opposite of what I find readable.

src/resources/js/docs.js
```diff
@@ -15,11 +17,17 @@ $(function() {
...
+		var $anchor = $(
+			'<a href="#' +
+				this.id +
+				'" class="anchor-link" title="Direct link">🔗</a>'
+		);
```
francislavoie: We can probably use a template literal instead here. I don't think we care about IE at this point lol, we can use ES6 stuff.

src/resources/js/docs.js
```diff
@@ -29,59 +37,120 @@ $(function() {
...
 			let text = item.innerText;
-			let url = text === '<directives...>'
-				? '/docs/caddyfile/directives'
-				: '/docs/caddyfile/directives/' + text;
-			text = text.replace(/</g,'&lt;').replace(/>/g,'&gt;');
-			$(item).html('<a href="' + url + '" style="color: inherit;" title="Directive">' + text + '</a>');
+			let url =
+				text === "<directives...>"
+					? "/docs/caddyfile/directives"
+					: "/docs/caddyfile/directives/" + text;
```
francislavoie, line 55: I also think putting code on the next line like this is worse. *shakes fist at Prettier*

src/resources/js/docs.js
```diff
@@ -29,59 +37,120 @@ $(function() {
...
+		let text = item.innerText.replace(/</g, "&lt;").replace(/>/g, "&gt;");
+		$(item).html(
+			'<a href="/docs/caddyfile/matchers#syntax" style="color: inherit;" title="Matcher token">' +
+				text +
+				"</a>"
+		);
+	});
+
+	// select all bash code elements and dynamically add clipboard button/svg
+	$("pre > code.bash").map(function (_, block) {
```
francislavoie: Ah, you only added it to `code.bash`, I see. I think in some cases it's `code.cmd > span.bash` as well. Or something like that.

src/resources/js/docs.js
danditomaso: I agree, I had realized prettier auto formatted it after I had created the PR for the work. I add prettier rc and going forwards it would avoid furher formatting issues.

danditomaso: I'm not a fan of some of the choices Prettier makes either, the new line for closing brackets is definitely one of them and its pretty annoying. I'll add a prettier Rc to ensure we're using a consistant set of formatting rules.

src/resources/js/docs.js
danditomaso: 100% and for anyone still using IE, they should really re-evaluate their browser life choices.  Thanks for the suggestion, I've added the change

src/resources/js/docs.js
danditomaso: Good catch! I've reduced the specificity of the selector and added span.bash.

src/resources/js/docs.js
danditomaso: Yeah its enabled in my VSCode, sorry about that. I do find it helpful for keeping CSS/JS well formatted (with cavets). I should have checked with you guys first to see if it was an accepted practice. I'm onboard with your opinion that some of the prettier rules dont make much sense. I'll create a second PR with some sane Prettier standards so we can avoid further formatting issues going forwards.

src/resources/js/docs.js
mholt: Ha, yeah, at this point I'm only concerned about the major browser engines :100:

src/resources/js/docs.js
mholt: I'm so glad you understand how to use this stuff :sweat_smile:

src/resources/css/docs.css
danditomaso: Thanks, I'm happy to add a note here, do you know what note I should add? It looks like more general selectors are above and more specific ones are below, but I'm guessing.

danditomaso:
> Hmm, this had some side-effects on `/docs/install`. For example, the Debian section's commands have a lot more padding now.
> 
> Before:
> 
> image
> 
> After:
> 
> image
> 
> Also I'm only seeing the copy button on some of the blocks on this page. I'm not sure why though, they aren't significantly different other than missing the button 🤔 Probably just a selector bug.

@francislavoie I'm seeing the same thing, I've added some changes to fix that in other locations. I think a better fix for this is for any case where there is a `code` element along with more than one span, is to add an inner div, keep the cmd's within that inner element along with some padding. Then absolutely position the copy button outside of that code block, which will ensure it wont overlap any text.  I wont have a chance to update this PR with thos fixes until late afternoon today if that is alright?

francislavoie:
No rush at all 👍 take your time

mholt:
Yeah, thanks for working on that! No problem, we're not in a hurry :)

mholt (approved):
Thank you for working on this, Dan! It looks good to me. :) Should be a great boon to our readers.

mholt:
Is this ready to merge? If not, feel free whenever it's ready :)

mohammed90:
@danditomaso, did you mean to close this? Is it not ready? Per the last feedback from Matt, we were just waiting on your confirmation that it's good to go.

PR #341, opened 2023-08-20
Lots of minor adjustments

francislavoie:
- Adjusted the `h*` CSS font sizes (making them all a bit bigger) to give room for `h6` to fit, it was way too small before because it inherited from the regular font size.
- Better explain placeholder behaviour for proxy addresses
- More root CA cert installation details and instructions in a bunch of places
- Move global opts in Concepts to earlier; it has higher importance now than it did in v2.0 so mentioning it earlier makes more sense
- More detail for site addresses
- Mention that LE cert expiry emails might be misleading
- Mention that on-demand should be fast, do simple DB lookups
- Deleted an old spec.md file that was never finished, most of it is outdated now and it seems essentially covered by the Concepts page anyway
- Reorder articles in nav, moved v2 upgrade to the bottom because lol it's a bit old now, and Architecture & Running make sense a bit higher in terms of importance


mholt (approved):
Sorry this took me a while. I appreciate your care over the docs! LGTM.

PR #342, opened 2023-08-22
Fix query filter typo for uri example

pauljeannot:
A very small typo which confused me a bit when starting with caddy.
This way the doc is more consistent with other examples mentionned around.


francislavoie:
Thanks!

PR #346, opened 2023-09-12
Add documentation for trusted_proxies_strict

nebez:
This is related to https://github.com/caddyserver/caddy/pull/5805

Adds documentation on how to enable strict mode, and what the implications of enabling it are on `client_ip`.


mholt:
Thank you so much for enhancing the docs and helping us maintain them!

nebez:
Cheers.

Once you and the Caddy team decide on which version to include this feature in, I will also update this PR to include that version in the documentation. Something along the lines of:

> As of v2.7.5, you can use option `trusted_proxies_strict` to do blahblahblah

francislavoie:
We're getting ready for the Caddy 2.8 release if you'd like to rebase and do some final polish!

nebez:
Great news! Rebased, introduced a reference to the version requirement, re-worded, and pointed PR to `v2.8` instead of `master`.

mholt (approved):

src/docs/markdown/caddyfile/options.md
```diff
@@ -822,7 +822,7 @@ Also included is the [`proxy_protocol`](/docs/json/apps/http/servers/listener_wr
...
+Enabling this causes trusted requests to have the _real_ client IP parsed from HTTP headers (by default, `X-Forwarded-For`; see [`client_ip_headers`](#client-ip-headers) to configure other headers). If trusted, the client IP is added to [access logs](/docs/caddyfile/directives/log), is available as a `{client_ip}` [placeholder](/docs/caddyfile/concepts#placeholders), and allows the use of the [`client_ip` matcher](/docs/caddyfile/matchers#client-ip). If the request is not from a trusted proxy, then the client IP is set to the remote IP address of the direct incoming connection. By default, [`client_ip_headers`](#client-ip-headers) are parsed left-to-right. See [`trusted_proxies_strict`](#trusted-proxies-strict) to alter this behaviour.
```
francislavoie: I think we can write it like this, we don't need to repeat `client_ip_headers`, it makes it misleading what this is trying to say, I think. My initial thought was "it's reading the config for `client_ip_headers` left to right" but it actually means "the IPs in those headers in the request".

```suggestion
Enabling this causes trusted requests to have the _real_ client IP parsed from HTTP headers (by default, `X-Forwarded-For`; see [`client_ip_headers`](#client-ip-headers) to configure other headers). If trusted, the client IP is added to [access logs](/docs/caddyfile/directives/log), is available as a `{client_ip}` [placeholder](/docs/caddyfile/concepts#placeholders), and allows the use of the [`client_ip` matcher](/docs/caddyfile/matchers#client-ip). If the request is not from a trusted proxy, then the client IP is set to the remote IP address of the direct incoming connection. By default, the IPs in the headers are parsed left-to-right. See [`trusted_proxies_strict`](#trusted-proxies-strict) to alter this behaviour.
```

src/docs/markdown/caddyfile/options.md
```diff
@@ -851,6 +851,27 @@ Here's a complete example, trusting an example IPv4 range and an IPv6 range:
 }
 ```
 
+##### `trusted_proxies_strict`
+
+When [`trusted_proxies`](#trusted-proxies) is enabled, the [`client_ip_headers`](#client-ip-headers) are parsed from left-to-right by default. The first untrusted IP address found becomes the real client address. Since v2.8, you can opt-in to right-to-left parsing of these headers with `trusted_proxies_strict`. By default, this option is disabled for backwards compatibility.
```
francislavoie: Same as above, but we can add the precision of which headers, cause this paragraph has no other mention of that config option unlike above.

```suggestion
When [`trusted_proxies`](#trusted-proxies) is enabled, the IPs in the headers (configured by [`client_ip_headers`](#client-ip-headers)) are parsed from left-to-right by default. The first untrusted IP address found becomes the real client address. Since v2.8, you can opt-in to right-to-left parsing of these headers with `trusted_proxies_strict`. By default, this option is disabled for backwards compatibility.
```

src/docs/markdown/caddyfile/options.md
nebez: Ah, great call. The wording does suggest the config is read left-to-right, instead of the values.

src/docs/markdown/caddyfile/options.md
nebez: Good call-out

francislavoie:
Thanks!

PR #349, opened 2023-10-25
docs: Add default for reverse_proxy health_timeout

steffenbusch:
Currently, https://caddyserver.com/docs/caddyfile/directives/reverse_proxy#active-health-checks is missing the default value for `health_timeout`. It has been added based on https://github.com/caddyserver/caddy/blob/d949caf459c3e046e6f4020dbb685c07812d5b80/modules/caddyhttp/reverseproxy/healthchecks.go#L88


mholt (approved):
Ah right, thank you!

PR #350, opened 2023-10-28
build: Suggest restarting service to load custom modules

wchargin:
I followed these instructions on a fresh Debian 12 box, seemingly
successfully. But `systemctl reload caddy` would reject my Caddyfile
that tried to use those custom modules, logging "unknown module" errors,
even though `caddy list-modules` reported that they were installed.
After reading [a comment on `caddyserver/cache-handler`][c], I tried
running `systemctl restart caddy`, which immediately fixed the issue.

In retrospect, it makes sense that since `apt install caddy` starts the
systemd service, replacing the binary on the filesystem does nothing to
change the running server. But the instructions didn't suggest this, and
I'm used to only `reload`ing Caddy, so I didn't think of it myself.

[c]: https://github.com/caddyserver/cache-handler/issues/33#issuecomment-1234457086

wchargin-branch: custom-build-systemctl-restart


mholt (approved):
I suppose this is a good idea; any objections @francislavoie ?

PR #362, opened 2023-12-19
running: mention systemd's `EnvironmentFile` directive

mohammed90:
Its benefit came up as part of the discussion in caddyserver/caddy#5987


francislavoie:
Thanks!

I think maybe we should recommend a real path to use though. Something like `/etc/caddy/.env`?

PR #365, opened 2024-01-01
Update install.md

Its-Just-Nans:
Add curl


mholt (approved):
I thought this already came installed on Debian/Ubuntu, but, I suppose you encountered a situation where it wasn't?

Its-Just-Nans:
Exact

you can see for example on debian it's not installed by default
```sh
# to test quickly with docker
docker pull debian
docker run -it debian /bin/bash
# try to use curl
# bash: curl: command not found
```

PR #370, opened 2024-02-01
restructure the "build from source" page

mohammed90:
I hope this is less confusing for the readers. I tried to make it clear that it's one-of-two ways that they need to follow, not both.

### DIFF
--- a/src/docs/markdown/build.md
+++ b/src/docs/markdown/build.md
@@ -4,10 +4,10 @@ title: "Build from source"
 
 # Build from source
 
-There are 2 ways to build Caddy, as opposed to installation of pre-built executable:
-- Build from Git repo
-- Build using `xcaddy`
-- Build custom Docker image
+There are 2 ways to build Caddy if you need to customize Caddy:
+- [Git](#git): Build from Git repo
+- [`xcaddy`](#xcaddy): Build using `xcaddy`
+- [Docker](#docker) Build custom Docker image
 
 Requirements:
 
@@ -15,7 +15,7 @@ Requirements:
 
 The [last section](#package-support-files-for-custom-builds-for-debianubunturaspbian) contains instructions for users who installed Caddy using the APT command on Debian-derivative system yet need the custom build executable for their operations.
 
-## Build from Git repo
+## Git
 
 Requirements:
 
@@ -49,7 +49,7 @@ Or similarly for Linux ARMv6 when you're not on Linux or on ARMv6:
 
 <pre><code class="cmd bash">GOOS=linux GOARCH=arm GOARM=6 go build</code></pre>
 
-## Build using `xcaddy`
+## xcaddy
 
 The [`xcaddy` command](https://github.com/caddyserver/xcaddy) is the easiest way to build Caddy with version information and/or plugins.
 
@@ -76,23 +76,9 @@ Cross-platform compilation with `xcaddy` works the same as with the `go` command
 
 <pre><code class="cmd bash">GOOS=darwin xcaddy build</code></pre>
 
-## Build Custom Docker Image
+## Docker
 
-Most users deploying production sites will not want to rely on mounting files into a container, but will instead base their own images on `caddy`:
-
-```Dockerfile
-# note: never use the :latest tag in a production site
-FROM caddy:<version>
-
-COPY Caddyfile /etc/caddy/Caddyfile
-COPY site /srv
-```
-
-#### Adding custom Caddy modules
-
-Caddy is extendable through the use of "modules". See https://caddyserver.com/docs/extending-caddy for full details. You can find a list of available modules on [the Caddy website's download page](https://caddyserver.com/download).
-
-You can use the `:builder` image as a short-cut to building a new Caddy binary:
+You can use the `:builder` image as a short-cut to building a new Caddy binary with custom modules:
 
 ```Dockerfile
 FROM caddy:<version>-builder AS builder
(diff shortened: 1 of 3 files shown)


src/docs/markdown/build.md
```diff
@@ -4,10 +4,24 @@ title: "Build from source"
 
 # Build from source
 
+There are 2 ways to build Caddy, as opposed to installation of pre-built executable:
```
francislavoie: > as opposed to installation of pre-built executable

I think we should change this to something like "if you need to customize Caddy" or something like that. "as opposed to" sounds negative

src/docs/markdown/build.md
```diff
@@ -4,10 +4,24 @@ title: "Build from source"
...
+- Build from Git repo
+- Build using `xcaddy`
```
francislavoie: Let's linkify these?

Also maybe we should add a `(recommended)` note for `xcaddy` here? Vast majority should be using `xcaddy` instead of from source.

src/docs/markdown/build.md
```diff
@@ -4,10 +4,24 @@ title: "Build from source"
...
+
 Requirements:
 
 - [Go](https://golang.org/doc/install) 1.20 or newer
 
+You can easily cross-compile, i.e. build Caddy on one operating system or architecture for a different, using the same tooling. You'll find details on this in the respective section.
+
+The latest section contains instructions for users who installed Caddy using the APT command on Debian-derivative system yet need the custom build executable for their operations.
```
francislavoie: "latest" usually feels like it relates to time rather than position in an article. "last" is usually better for that.

This should also probably link to the section below.

src/docs/markdown/build.md
```diff
@@ -4,10 +4,24 @@ title: "Build from source"
...
 
+There are 2 ways to build Caddy, as opposed to installation of pre-built executable:
+- Build from Git repo
+- Build using `xcaddy`
+
 Requirements:
 
 - [Go](https://golang.org/doc/install) 1.20 or newer
 
+You can easily cross-compile, i.e. build Caddy on one operating system or architecture for a different, using the same tooling. You'll find details on this in the respective section.
```
francislavoie: Unnecessary note I think

```suggestion
You can easily cross-compile, i.e. build Caddy on one operating system or architecture for a different, using the same tooling.
```

src/docs/markdown/build.md
```diff
@@ -4,10 +4,24 @@ title: "Build from source"
 
 # Build from source
 
+There are 2 ways to build Caddy, as opposed to installation of pre-built executable:
+- Build from Git repo
+- Build using `xcaddy`
+
```
francislavoie, line 11: I think we should probably add a section about building with Docker in here. It's a quite common need, no reason we don't just have this in our docs. We have it in https://hub.docker.com/_/caddy but that's not where people usually look I think. See the "Adding custom Caddy modules" section there.

src/docs/markdown/build.md
```diff
@@ -26,8 +40,17 @@ Due to [a bug in Go](https://github.com/golang/go/issues/29228), these basic ste
...
+For example, to compile Caddy for Windows when you're not on Windows:
+
+<pre><code class="cmd bash">GOOS=windows go build</code></pre>
+
+Or similarly for Linux ARMv6 when you're not on Linux or on ARMv6:
+
+<pre><code class="cmd bash">GOOS=linux GOARCH=arm GOARM=6 go build</code></pre>
 
-## xcaddy
+## Build using `xcaddy`
```
francislavoie: I think changing this header will break a bunch of old `#xcaddy` anchor links in the forums etc.

Something we can do is add a tiny little `<script>` at the top of the page which just does like:
```html
<script>
if (window.location.hash == '#xcaddy') window.location.hash = '#build-using-xcaddy';
</script>
```

Would redirect the user.

src/docs/markdown/build.md
mohammed90: How about:

```
- [Git](#git): Build from Git repo
- [`xcaddy`](#xcaddy): Build using `xcaddy` (Recoomended)
```

? I'll change the section headers to `Git` and `xcaddy`

src/docs/markdown/build.md
mohammed90: If the approach proposed in the [comment above](https://github.com/caddyserver/website/pull/370#discussion_r1474056196) is okay, there will be no breaking links. I can include this code snippet if the section ID ends up being differnt.

src/docs/markdown/build.md
mohammed90: I copy-pasted that section from Docker Hub into a section on this page. Good?

src/docs/markdown/build.md
francislavoie: Sure 👍

src/docs/markdown/build.md
```diff
@@ -50,24 +72,45 @@ To build with plugins, use `--with`:
...
 
-For example, to compile Caddy for Windows when you're not on Windows:
+```Dockerfile
+# note: never use the :latest tag in a production site
+FROM caddy:<version>
 
-<pre><code class="cmd bash">GOOS=windows go build</code></pre>
+COPY Caddyfile /etc/caddy/Caddyfile
+COPY site /srv
+```
```
francislavoie: Let's skip this part, not relevant to "building Caddy" really.

src/docs/markdown/build.md
```diff
@@ -50,24 +72,45 @@ To build with plugins, use `--with`:
...
 
-Or similarly for Linux ARMv6 when you're not on Linux or on ARMv6:
+#### Adding custom Caddy modules
 
-<pre><code class="cmd bash">GOOS=linux GOARCH=arm GOARM=6 go build</code></pre>
+Caddy is extendable through the use of "modules". See https://caddyserver.com/docs/extending-caddy for full details. You can find a list of available modules on [the Caddy website's download page](https://caddyserver.com/download).
```
francislavoie: I think this line belongs elsewhere than this section, it only makes sense in context of the Docker hub docs which needs to give some background/preamble

src/docs/markdown/build.md
```diff
@@ -50,24 +72,45 @@ To build with plugins, use `--with`:
...
+    --with github.com/hairyhenderson/caddy-teapot-module@v0.0.3-0
+
+FROM caddy:<version>
+
+COPY --from=builder /usr/bin/caddy /usr/bin/caddy
+```
+
+Note the second `FROM` instruction - this produces a much smaller image by simply overlaying the newly-built binary on top of the regular `caddy` image.
+
+The [`xcaddy`](https://caddyserver.com/docs/build#xcaddy) tool is used to [build a new Caddy entrypoint](https://github.com/caddyserver/caddy/blob/4217217badf220d7d2c25f43f955fdc8454f2c64/cmd/caddy/main.go#L15..L25), with the provided modules. You can specify just a module name, or a name with a version (separated by `@`). You can also specify a specific version (can be a version tag or commit hash) of Caddy to build from. Read more about [`xcaddy` usage](https://github.com/caddyserver/xcaddy#command-usage).
```
francislavoie: Can simplify links to just anchor since they're on this page. I don't think we need to mention the entrypoint stuff or xcaddy syntax here, it should just link to the above section I think.

src/docs/markdown/build.md
```diff
@@ -50,24 +72,45 @@ To build with plugins, use `--with`:
...
+
+Note that the "standard" Caddy modules ([`github.com/caddyserver/caddy/master/modules/standard`](https://github.com/caddyserver/caddy/tree/master/modules/standard)) are always included.
```
francislavoie: I think this is redundant too.

francislavoie:
I did some cleanup, added more cross-linking, fixed a couple random bugs I noticed (metrics page looked weird).

LGTM now, feel free to merge it if you're happy with it @mohammed90!

francislavoie (approved):

PR #371, opened 2024-02-06
add ids to some sections on features page

mohammed90:
There's a need to link to the "Automatic HTTPS" and "ACME" sections. I've added an "id" attribute so we can link by fragment.


mholt (approved):
Good idea. Thanks!

PR #372, opened 2024-02-08
Switch to jquery 4.0

francislavoie:
https://blog.jquery.com/2024/02/06/jquery-4-0-0-beta/

Any objections @mholt? Seems to work fine from some quick clicking around.


mholt (approved):
Yeah, that should be fine I think

Long live jQuery

Issue #364, opened 2023-12-30
`time_format` "compatible time layout string" could do with clarification


polarathene:
## Problem

https://caddyserver.com/docs/caddyfile/directives/log#format-modules

The last option listed for `time_format` is:

> _Or, any compatible time layout string; see the [Go documentation](https://pkg.go.dev/time#pkg-constants) for full details._

For those that are Go developers perhaps those instructions are more obvious? (_this was [recognized as a documentation concern in Feb 2022](https://caddy.community/t/questions-about-logs/15160/7)_)

What the user sees after visiting the link (_note the scrollbar size and all the additional content on this page, when all the Caddy user wants is to know what values they can use_):

![image](https://github.com/caddyserver/website/assets/5098581/f34b75be-f0b0-4283-8025-35a404198418)

---

## UX

Perhaps there is more context relevant in the document, but the page itself is rather large.

After messing around with it for a while and looking up some discussions from the Caddy community forum threads, it seems that you're meant to copy/paste the string value assigned to any of those constants.

`TimeOnly   = "15:04:05"` for example requires equivalent Caddyfile:

```Caddyfile
log {
  format console {
    time_format "15:04:05"
  }
}
```

This wasn't obvious as adjusting those values does appear to affect the time. `"15:01:05` for example renders the minutes as fixed value of `12`, while `02` was fixed at `29` and `00` fixed at `00`? 😕 

Initially I thought the docs might have been referring to the constant names (_this just outputs the value as a string, not a timestamp_), or even this section:

![image](https://github.com/caddyserver/website/assets/5098581/e577ee7a-02a4-43ce-9c3c-e54a050b2460)

`"-07:00:00"` outputs `+00:00:00` and `"Z07:00:00"` outputs `Z`.

Likewise attempts with [format string syntax documented will just output those as a string](https://caddy.community/t/questions-about-logs/15160/6) instead of formatting the time like one might expect was implied.

---

## Fix (Include some examples / context)

Caddy docs could more clearly communicate that a string value from the constants should be used.

Ideally providing an example or two to further strengthen the referenced content and usage.

Add those to the [log docs examples section](https://caddyserver.com/docs/caddyfile/directives/log#examples), and have the `time_format` option reference it? (_unfortunately individual examples can't be linked, so inlining the example may be better to co-locate it instead?_)

Something along the lines of

```Caddyfile
# NOTE: Do not adjust the layout string value, it is not a format example.

# Use the "TimeOnly" layout string (hh:mm:ss):
[...]

PR #374, opened 2024-02-16
Big Caddyfile docs update

francislavoie:
There's a lot here.

Mostly, improved examples. In most cases, I switched to using the full Caddyfile syntax with a site block instead of simple directives. We got feedback from users that were confused where things go in the Caddyfile when examples aren't in-context.


francislavoie:
I'm not done yet, I need to make more improvements to the global options page. Most other pages are in a solid place tho.

src/docs/markdown/caddyfile/directives/log.md
```diff
@@ -171,33 +199,43 @@ format <encoder_module> {
...
+  - `unix_milli_float` Floating-point number of milliseconds since the Unix epoch.
+  - `unix_nano` Integer number of nanoseconds since the Unix epoch.
+  - `iso8601` Example: `2006-01-02T15:04:05.000Z0700`
+  - `rfc3339` Example: `2006-01-02T15:04:05Z07:00`
+  - `rfc3339_nano` Example: `2006-01-02T15:04:05.999999999Z07:00`
+  - `wall` Example: `2006/01/02 15:04:05`
+  - `wall_milli` Example: `2006/01/02 15:04:05.000`
+  - `wall_nano` Example: `2006/01/02 15:04:05.000000000`
+  - `common_log` Example: `02/Jan/2006:15:04:05 -0700`
   - Or, any compatible time layout string; see the [Go documentation](https://pkg.go.dev/time#pkg-constants) for full details.
```
polarathene, line 240: Not sure if you want to bundle addressing https://github.com/caddyserver/website/issues/364 in this PR as well, since you've got so many changes already going on 😅 

Unlike the other values above, you're meant to provide the fixed value assigned (similar to that of the "Example" values above this line in the docs.

```
# NOTE: Do not adjust the layout string value, it is not a format example.

# Use the "TimeOnly" layout string (hh:mm:ss):
log {
  format console {
    time_format "15:04:05"
  }
}

# Use the "DateOnly" layout string (yyyy-mm-dd):
log {
  format console {
    time_format "2006-01-02"
  }
}
```

As a user (and not a Go dev) I recall that taking a while to grok. Especially with the value being accepted when altered, but to actually get the expected functionality must be left as-is.

For context, I was only interested in `TimeOnly` / `Kitchen` for console logs to avoid noise from the timestamp. The ones Caddy offered with friendly names had nothing as terse on offer.

src/docs/markdown/caddyfile/options.md
```diff
@@ -406,6 +689,8 @@ For example, to configure different options for the servers on ports `:80` and `
 }
 ```
 
+When using `servers`, it will **only** apply to servers that **actually appear** in your Caddyfile (i.e. are produced by a site block). Remember, [Automatic HTTPS](/docs/automatic-https) will create a server listening on port `80`, to serve HTTP->HTTPS redirects and to solve the ACME HTTP challenge; this happens at runtime, i.e. _after_ the Caddyfile adapter applies `servers`. So in other words, this means that `servers` **will not** apply to `:80` unless you explicitly declare a site block like `http://` or `:80`.
```
polarathene: ❤️ 

Is port 80 correct here, or is it actually tied to the `http_port` setting? (_likewise for other related content making this reference to port 80_)

src/docs/markdown/caddyfile/options.md
```diff
@@ -472,26 +756,34 @@ http:// {
...
+		listener_wrappers {
+			http_redirect
+			tls
+		}
+	}
 }
 ```
 
-Also included is the [`proxy_protocol`](/docs/json/apps/http/servers/listener_wrappers/proxy_protocol/) listener wrapper (prior to v2.7.0 it was only available via a plugin), which enables [PROXY protocol](https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt) parsing (popularized by HAProxy). This must be used _before_ the `tls` listener wrapper since it parses plaintext data at the start of the connection:
+Also included is the [`proxy_protocol`](/docs/json/apps/http/servers/listener_wrappers/proxy_protocol/) listener wrapper (prior to v2.7.0 it was only available via a plugin), which enables [PROXY protocol](https://github.com/haproxy/haproxy/blob/master/doc/proxy-protocol.txt) parsing (popularized by HAProxy). This must be used _before_ the `tls` listener wrapper since it parses plaintext data at the start of the connection:
```
polarathene, line 774: 👍

src/docs/markdown/caddyfile/options.md
```diff
@@ -472,26 +756,34 @@ http:// {
...
-	proxy_protocol {
-		timeout 2s
-		allow 192.168.86.1/24 192.168.86.1/24
+```caddy
+{
+	servers {
+		listener_wrappers {
+			proxy_protocol {
+				timeout 2s
+				allow 192.168.86.1/24 192.168.86.1/24
```
polarathene, line 782: With Caddy 2.8 this would also have a deny setting, but perhaps you want to delay adding that until release? I don't think Caddy offers version docs.

2.7 vs 2.8 has another notable change with 2.8 allowing connections without a PROXY protocol header through. Unless it changes to `REQUIRE` as a default (configurable?) policy for `allow`?

src/docs/markdown/caddyfile/directives/basicauth.md
```diff
@@ -37,11 +37,30 @@ basicauth [<matcher>] [<hash_algorithm> [<realm>]] {
 
 ## Examples
 
-Protect all resources in /secret so only Bob can access them with the password "hiccup":
+Require authentication for all requests to `example.com`:
+
+```caddy
+example.com {
+	basicauth {
```
polarathene, line 44: I recall seeing `basicauth` being changed to `basic_auth` for Caddy 2.8 too.

I guess this feedback is better deferred to a follow-up PR (_or already planned_)

polarathene (commented):
Just a small review with feedback for the PROXY protocol changes.

---

Fantastic work with the overall improvements to docs though! 😎 

I have definitely been confused by how to interpret the Caddyfile docs in the past, the added clarity is awesome 🎉

src/docs/markdown/caddyfile/options.md
francislavoie: Yeah, avoiding 2.8 changes for now. We'll do another pass for that closer to release (and merge on release)

src/docs/markdown/caddyfile/directives/basicauth.md
francislavoie: Yep, deferred for 2.8

src/docs/markdown/caddyfile/directives/log.md
francislavoie: > Note that the parts of the format string are special constants for the layout; so `2006` is the year, `01` is the month, `Jan` is the month as a string, `02` is the day. Do not use the actual current date numbers in the format string.

src/docs/markdown/caddyfile/options.md
francislavoie: I was being lazy there :+1:

PR #378, opened 2024-03-11
Update acme_eab in options.md

PhilippKolmann:
I wanted to use ACME_EAB in global config but failed to do so with the config.

After investigating in the source, I saw that it's configured differently to the acme issuer eab config block.

### DIFF
--- a/src/docs/markdown/caddyfile/options.md
+++ b/src/docs/markdown/caddyfile/options.md
@@ -86,7 +86,10 @@ Possible options are (click on each option to jump to its documentation):
 	skip_install_trust
 	acme_ca <directory_url>
 	acme_ca_root <pem_file>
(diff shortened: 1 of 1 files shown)


francislavoie:
Ah, we should probably also fix it to work inline as well.

mholt:
Oops. Well, I'm good with either way (we can either fix the docs or fix the code. I don't have a strong preference as to which). With EAB, usually the key ID is expressed before the hmac so the order is somewhat intuitive, even if unlabeled. But the separate lines with the labels are much more clear/obvious.

francislavoie:
I just mean both should work, but I think it's fine to make this docs change now. I'll merge. If you'd like to make a PR to adjust the Caddy code to allow both inline and block config, that'd be great :thumbsup:

PR #380, opened 2024-03-21
fix keepalive_interval link at reverse_proxy

x86pup:
The config option hyperlink for `keepalive_interval` is broken (https://caddyserver.com/docs/caddyfile/directives/reverse_proxy#the-http-transport), and I'm guessing this is why.


francislavoie:
Ah yep, thanks, lazy copy-paste :grin:

PR #386, opened 2024-03-30
Update install.md - Add OPNsense installation method

Monviech:
I'm not sure if a more specialized implementation of Caddy is also allowed to be shown on the website. This integration offers a GUI inside the OPNsense Firewall.

The os-caddy package is the GUI component, and it depends on caddy-custom, an xcaddy built Caddy in the FreeBSD ports.

Since OPNsense runs on FreeBSD, Caddy can be used like on any other plattform (by editing the Caddyfile and not using the provided GUI if wanted). 

I think it's worth a shot to offer a quick example of this community maintained installation method, since it's a part of OPNsense ports and plugins now.


mholt (approved):
Awesome, yeah that is great :) Thank you for the contribution!

PR #387, opened 2024-04-04
Fix prometheus.yml metrics config

dylanschultzie:
Prometheus metrics require the `/metrics` path to work correctly.

### DIFF
--- a/src/docs/markdown/metrics.md
+++ b/src/docs/markdown/metrics.md
@@ -48,6 +48,7 @@ global:
 
 scrape_configs:
   - job_name: caddy
+    metrics_path: /metrics
     static_configs:
       - targets: ['localhost:2019']
 ```


hairyhenderson:
@dylanschultzie that's [the default](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config), so it doesn't need to be specified.

IMO this shouldn't be added as it isn't necessary and complicates matters. If someone needs to know how to configure Prometheus, they should read their docs (as referenced above).

dylanschultzie:
@hairyhenderson I don't agree with excluding it. Using Prometheus v2.48.0, I had to specify `/metrics` for them to be consumed.

That being the case, I'd argue omitting it being more detrimental than not. In addition, if `caddy` is already providing a barebones scrape config, why not ensure it's correct?

francislavoie:
Re-stating the defaults doesn't make sense. 

What's your evidence that it doesn't work without it?

dylanschultzie:
Listen, I'm just trying to help out. 

Potentially forcing users to dig through community forums, or the Prometheus docs "(as referenced above)" to find pertinent information does nobody any good when it can be solved simply in a single line.

francislavoie:
You haven't shown evidence that it actually solves a problem. We don't make changes to the docs without evidence.

PR #388, opened 2024-04-08
Fix broken link for otel env variables

jtmiclat:
The following page https://caddyserver.com/docs/caddyfile/directives/tracing directs to https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md which returns a 404. 

I updated it to https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md)

Let me know if it would be better to reference a tagged version so it won't break in the future!


mholt (approved):
Thank you!

No idea about whether a tagged version is better but this is at least good for now. Will let another maintainer merge if they agree.

PR #400, opened 2024-06-13
snippet block input documentation

elee1766:
i added experimental + the tentative version tag. 

i think it would be good to make sure people know such a feature is 2.x+ only, as I think most people will read the snippet page, and might get confused if running an old version.

### DIFF
--- a/Caddyfile
+++ b/Caddyfile
@@ -1,4 +1,4 @@
-localhost
+http://localhost:2080
 
 root * src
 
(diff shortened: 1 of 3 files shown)


francislavoie (commented):
Thanks! Looks good.

I think we could also add a note about this on the Caddyfile Concepts page in the snippets section. 

I think you have some mixed tabs and spaces in the example, tabs are preferred for consistency.

elee1766:
both should be done.

mholt:
Thanks so much for working on this!

PR #401, opened 2024-06-15
reverse_proxy: add health_follow_redirects

fin444:
Documents the new `health_follow_redirects` setting added in https://github.com/caddyserver/caddy/pull/6302

As an aside, this change probably should have been mentioned in v2.8.0's release notes since it effectively breaks existing health checks that rely on redirects.


francislavoie:
Thanks! Sorry, we're behind on the docs updates for this release, we all are having a busy end of spring.

You were relying on redirects? That's surprising. What's your usecase?

fin444:
> You were relying on redirects? That's surprising. What's your usecase?

I set up health checks pointing to my sites, and they were working on 2.7. Wasn't paying attention to whether they were redirects or not :)

PR #403, opened 2024-06-20
docs: remove outdated zstd browser support comment

lucasmirloup:
zstd browser support : https://caniuse.com/zstd

### DIFF
--- a/src/docs/markdown/caddyfile-tutorial.md
+++ b/src/docs/markdown/caddyfile-tutorial.md
@@ -139,11 +139,6 @@ templates
 file_server browse
 ```
 
-<aside class="tip">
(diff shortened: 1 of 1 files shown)


mholt (changes requested):
Thanks for the PR!

Actually, according to the link, all serious mainstream browsers support Zstandard now. (Safari is... erm, special... I don't consider it a "serious" browser at this point given how slow they are to actually implement standards among other reasons.)

Since pretty much every browser supports it, maybe we should just remove this section completely.

lucasmirloup:
OK, I've updated the PR :+1:

mholt (approved):

mholt:
Thanks!

PR #404, opened 2024-06-29
Explain what is needed to delegate a subdomain to another zone

pimlie:
In the DNS Challenge chapter it is said that _This can be used to delegate the \_acme-challenge subdomain to another zone_, but it is not further explained _how_ you can configure that.

So I propose to add a link to the `dns_challenge_override_domain` setting which users need to configure if they wish to resolve the acme_challenge for domain1 through the zone of domain2.


francislavoie (approved):
Thanks!

PR #409, opened 2024-07-25
developer documentation for placeholders

elee1766:
also adds a docker command to the readme that might help people working with this repo

i run caddy already locally, so it's a bit of a pain to have to disable admin. (also i cant bind to 80/443 without root)

hosting it here
https://caddy-website.put.gay/docs/extending-caddy/placeholders

### DIFF
--- a/src/docs/markdown/extending-caddy/placeholders.md
+++ b/src/docs/markdown/extending-caddy/placeholders.md
@@ -4,9 +4,9 @@ title: "Placeholder Support"
 
 # Placeholders
 
-In Caddy, placeholders are a feature of the individual plugins. They are not parsed at config time, but instead preserved, and replaced at runtime.
+In Caddy, placeholders are processed by each individual plugin themselves. They are not parsed at config time, but instead preserved and processed at runtime.
 
-This means that if you wish for your plugin to support placeholders, you must explicitly add support for it.
+This means that if you wish for your plugin to support placeholders, you must explicitly add support for them.
 
 If you are not yet familiar with placeholders, start by reading [here](/docs/conventions#placeholders)!
 
@@ -75,18 +75,18 @@ When you adapt this Caddyfile with `HOST=example caddy adapt` you will get
 
 Importantly, look at the `"body"` field in both `srv0` and `srv1`.
 
-Since `srv0` used `{$ENV}`, the special environmental variable replacement with `$` became `example`, as it is parsed during Caddyfile parse time.
+Since `srv0` used `{$HOST}`, the special environmental variable replacement with `$`, the value became `example`, as it was processed during Caddyfile parse time.
 
 Since `srv1` used `{env.HOST}`, a normal placeholder, it was parsed as its own raw string value, `{env.HOST}`
 
-This means that down the line, the handler the plugins within `srv0` and `srv1` will receive the raw values `example` and `{env.Host}` respectively in their configurations.
+Some users may immediately notice that this means it is impossible to use the `{$ENV}` syntax in a JSON config. The solution to this is to process such placeholders at Provision time, which is covered below.
 
 
-## How to Placeholders in your Plugin
+## How to use Placeholders in your Plugin
 
-#### Parse the raw Placeholder in your unmarshaler
+#### Parse the raw Placeholder value in your unmarshaler
 
-Placeholders should be parsed as their raw values when parsing the Cazddyfile, just like any other string value
+Placeholders should be parsed as their raw values when parsing the Cazddyfile, just like any other string value.
 
 ```go
 func (g *Gizmo) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
@@ -115,7 +115,7 @@ func (g *Gizmo) ServeHTTP(w http.ResponseWriter, r *http.Request, next caddyhttp
 
 #### Alternatively, resolve the Placeholder during Provision
 
-If you only use global placeholders, like `env`, then you may initialize a global replacer at provision time, and use it to replace such values.
+If you only use global placeholders, like `env`, then you may initialize a global replacer at provision time, and use it to replace such values. This also allows users of config file formats other than Caddyfile to use environmental variables.
 
 ```go
 func (g *Gizmo) Provision(ctx caddy.Context) error {
--- a/src/docs/markdown/extending-caddy/placeholders.md
+++ b/src/docs/markdown/extending-caddy/placeholders.md
@@ -12,7 +12,7 @@ If you are not yet familiar with placeholders, start by reading [here](/docs/con
 
 ## Placeholder Internals
 
-Internally, placeholders are simply a string in format `{foo.bar}` used as valid configuration values, which is later parsed at runtime.
+Placeholders are a string in the format `{foo.bar}` used as dynamic configuration values, which is later evaluated at runtime.
 
 Placeholders-like strings which start with a dollar sign (`{$FOO}`), are evaulated at Caddyfile parse time, and do not need to be dealt with by your plugin. This is because these are not placeholders, but Caddyfile-specific [environmental variable substitution](/docs/caddyfile/concepts#environment-variables), they just happen to share the `{}` syntax.
 
--- a/src/docs/markdown/extending-caddy/placeholders.md
+++ b/src/docs/markdown/extending-caddy/placeholders.md
@@ -10,7 +10,7 @@ This means that if you wish for your plugin to support placeholders, you must ex
 
 If you are not yet familiar with placeholders, start by reading [here](/docs/conventions#placeholders)!
 
-## Placeholder Internals
+## Placeholders Overview
 
 Placeholders are a string in the format `{foo.bar}` used as dynamic configuration values, which is later evaluated at runtime.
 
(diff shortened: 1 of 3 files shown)


src/docs/markdown/extending-caddy/placeholders.md
```diff
@@ -0,0 +1,138 @@
+---
+title: "Placeholder Support"
+---
+
+# Placeholders
+
+In Caddy, placeholders are processed by each individual plugin themselves. They are not parsed at config time, but instead preserved and processed at runtime.
```
mholt: > They are not parsed at config time, but instead preserved and processed at runtime.

Technically, this depends if they're evaluated during Provision or later on. Maybe more specfic:

```suggestion
In Caddy, placeholders are processed by each individual plugin as needed; they do not automatically work everywhere.
```

src/docs/markdown/extending-caddy/placeholders.md
```diff
@@ -0,0 +1,138 @@
...
+
+This means that if you wish for your plugin to support placeholders, you must explicitly add support for them.
+
+If you are not yet familiar with placeholders, start by reading [here](/docs/conventions#placeholders)!
+
+## Placeholder Internals
+
+Internally, placeholders are simply a string in format `{foo.bar}` used as valid configuration values, which is later parsed at runtime.
```
mholt:
```suggestion
Placeholders are a string in the format `{foo.bar}` used as dynamic configuration values, which is later evaluated at runtime.
```

src/docs/markdown/extending-caddy/placeholders.md
```diff
@@ -0,0 +1,138 @@
...
+
+Placeholders-like strings which start with a dollar sign (`{$FOO}`), are evaulated at Caddyfile parse time, and do not need to be dealt with by your plugin. This is because these are not placeholders, but Caddyfile-specific [environmental variable substitution](/docs/caddyfile/concepts#environmental-variables), they just happen to share the `{}` syntax.
+
+It is therefore important to understand that `{env.HOST}` is inherently different from something like `{$HOST}`
```
mholt:
```suggestion
It is therefore important to understand that `{env.HOST}` is inherently different from something like `{$HOST}`.
```

src/docs/markdown/extending-caddy/placeholders.md
```diff
@@ -0,0 +1,140 @@
...
+Importantly, look at the `"body"` field in both `srv0` and `srv1`.
+
+Since `srv0` used `{$HOST}`, the special environmental variable replacement with `$`, the value became `example`, as it was processed during Caddyfile parse time.
+
+Since `srv1` used `{env.HOST}`, a normal placeholder, it was parsed as its own raw string value, `{env.HOST}`
+
+Some users may immediately notice that this means it is impossible to use the `{$ENV}` syntax in a JSON config. The solution to this is to process such placeholders at Provision time, which is covered below.
+
+
+## How to use placeholders in your plugin
```
mholt, line 85:
```suggestion
## Supporting placeholders
```

src/docs/markdown/extending-caddy/placeholders.md
```diff
@@ -0,0 +1,140 @@
...
+
+## Placeholder Internals
+
+Internally, placeholders are simply a string in format `{foo.bar}` used as valid configuration values, which is later parsed at runtime.
+
+Placeholders-like strings which start with a dollar sign (`{$FOO}`), are evaulated at Caddyfile parse time, and do not need to be dealt with by your plugin. This is because these are not placeholders, but Caddyfile-specific [environmental variable substitution](/docs/caddyfile/concepts#environment-variables), they just happen to share the `{}` syntax.
+
+It is therefore important to understand that `{env.HOST}` is inherently different from something like `{$HOST}`
+
+As an example, see the following caddyfile:
```
mholt:
```suggestion
As an example, see the following Caddyfile:
```

src/docs/markdown/extending-caddy/placeholders.md
```diff
@@ -0,0 +1,140 @@
...
+Since `srv0` used `{$HOST}`, the special environmental variable replacement with `$`, the value became `example`, as it was processed during Caddyfile parse time.
+
+Since `srv1` used `{env.HOST}`, a normal placeholder, it was parsed as its own raw string value, `{env.HOST}`
+
+Some users may immediately notice that this means it is impossible to use the `{$ENV}` syntax in a JSON config. The solution to this is to process such placeholders at Provision time, which is covered below.
+
+
+## How to use placeholders in your plugin
+
+#### Parse the raw placeholder value in your unmarshaler
```
mholt: I don't really think this section adds any value / says anything interesting. It can probably be removed...

src/docs/markdown/extending-caddy/placeholders.md
```diff
@@ -0,0 +1,140 @@
...
+func (g *Gizmo) UnmarshalCaddyfile(d *caddyfile.Dispenser) error {
+	d.Next()
+	if !d.Args(&g.Name) {
+		// not enough args
+		return d.ArgErr()
+	}
+}
+```
+
+#### Evaluate the placeholder during Match or Serve
```
mholt: I would reframe this section and the next to be more general.

It boils down to whether a Replacer is already available or not. If one has already been made, as is the case with the HTTP handler chain, it can be obtained from the context. I think we can talk about it with that as the example, but less specific to "Match" or "ServeHTTP". The only real difference is how you get the Replacer: either from a context, or make a new one.

mholt (changes requested):
Thanks for taking the initiative on this! I think this will be a great addition to the docs. I have some thoughts for your review.

src/docs/markdown/extending-caddy/placeholders.md
elee1766: hmmm, my thought is that some people may be confused thinking they might need to use a special function to parse a placeholder.

if you don't think thats a problem, we can gut it

src/docs/markdown/extending-caddy/placeholders.md
elee1766: do you think it should be reframed to be more general, or there be instead something two articles/examples showing

1. explaining how to grab a new replacer, and using that to resolve a placeholder
2. explaining how to create a replacer from a context.Context.

with a blurb saying that to use placeholder in code, pass the string into a replacer?

elee1766:
@mholt accepted all the grammar changes, had two questions for you on your last two comments that will change what i write to replace the last section. i agree that it's weak.

src/docs/markdown/extending-caddy/placeholders.md
```diff
@@ -0,0 +1,140 @@
...
+
+# Placeholders
+
+In Caddy, placeholders are processed by each individual plugin as needed; they do not automatically work everywhere.
+
+This means that if you wish for your plugin to support placeholders, you must explicitly add support for them.
+
+If you are not yet familiar with placeholders, start by reading [here](/docs/conventions#placeholders)!
+
+## Placeholder Internals
```
mholt:
```suggestion
## Placeholder Overview
```

src/docs/markdown/extending-caddy/placeholders.md
```diff
@@ -0,0 +1,140 @@
...
+Importantly, look at the `"body"` field in both `srv0` and `srv1`.
+
+Since `srv0` used `{$HOST}`, the special environmental variable replacement with `$`, the value became `example`, as it was processed during Caddyfile parse time.
+
+Since `srv1` used `{env.HOST}`, a normal placeholder, it was parsed as its own raw string value, `{env.HOST}`
+
+Some users may immediately notice that this means it is impossible to use the `{$ENV}` syntax in a JSON config. The solution to this is to process such placeholders at Provision time, which is covered below.
+
+
+## How to use placeholders in your plugin
```
mholt:
```suggestion
## Implementing support for placeholders
```

src/docs/markdown/extending-caddy/placeholders.md
mholt: What if we remove the section but put what you're conveying in the parent section ("Implementing support for placeholders" as a heading I suggest):

> Placeholders are evaluated at runtime (not at Caddyfile parsing or config loading). When you are unmarshaling a Caddyfile, do not evaluate placeholders at that time. Placeholders may be evaluated during `Provision()` or during your module's execution.

src/docs/markdown/extending-caddy/placeholders.md
mholt: I think this one article is enough; mainly let's show examples of both in this one.

So show how to get it from the context, where it will often be stored, and also how to create a new one if there isn't one already available.

mholt (changes requested):
Thanks for the updates! Almost there. I hope I'm able to answer your questions alright. They are good questions, thanks for working on this!

src/docs/markdown/extending-caddy/placeholders.md
elee1766: i chose `Implementing placeholder support` to better match the article title `implementing placeholders`

elee1766:
@mholt i rewrote the last section into a single examples section with an intro - it feels a lot nicer. 

also rewrote it to be more instructional/imperative, as at this point, the user should have already read the overview, and just probably wants to know what to do.

only thought - should i also include an example of parsing a string? or should people be expected to know that.

src/docs/markdown/extending-caddy/placeholders.md
```diff
@@ -82,56 +82,30 @@ Since `srv1` used `{env.HOST}`, a normal placeholder, it was parsed as its own r
...
-		return err
-	}
-	return next.ServeHTTP(w, r)
-}
-```
-
-#### Alternatively, resolve the placeholder during Provision
-
-If you only use global placeholders, like `env`, then you may initialize a global replacer at provision time, and use it to replace such values. This also allows users of config file formats other than Caddyfile to use environmental variables.
+In this example, we are using a newly constructed replacer to process placeholders. It has access to placeholders such as `{env.HOST}`, but NOT `{http.request.uri}`
```
francislavoie:
```suggestion
In this example, we are using a newly constructed replacer to process placeholders. It has access to [global placeholders](/docs/conventions#placeholders) such as `{env.HOST}`, but NOT request placeholder such as `{http.request.uri}`
```

elee1766: i accepted this manually.

src/docs/markdown/extending-caddy/placeholders.md
```diff
@@ -0,0 +1,113 @@
...
+Since `srv0` used `{$HOST}`, the special environmental variable replacement with `$`, the value became `example`, as it was processed during Caddyfile parse time.
+
+Since `srv1` used `{env.HOST}`, a normal placeholder, it was parsed as its own raw string value, `{env.HOST}`
+
+Some users may immediately notice that this means it is impossible to use the `{$ENV}` syntax in a JSON config. The solution to this is to process such placeholders at Provision time, which is covered below.
+
+
+## Implementing placeholder support
+
+You should not process placeholders when ummarshaling your Caddyfile. Instead, unmarshal the placeholders as strings in your configuration and evaluate them during either your module's execution or `Provision()` using a `caddy.Replacer`.
```
francislavoie:
```suggestion
You should not process placeholders when ummarshaling your Caddyfile. Instead, unmarshal the placeholders as strings in your configuration and evaluate them during either your module's execution (e.g. `ServeHTTP()` for HTTP handlers, `Match()` for matchers, etc.) or in the `Provision()` step, using a `caddy.Replacer`.
```

francislavoie (approved):
After that last thing :point_up:, looks good to me, can merge.

src/docs/markdown/extending-caddy/placeholders.md
elee1766: reads well to me

elee1766:
done

mholt (approved):
Great! I think this looks better now. :) Thank you!

PR #414, opened 2024-09-04
docs: Update ZeroSSL issuer for v2.8

kekalainen:
Addresses the changes in caddyserver/caddy#6229 to the best of my understanding.

Particularly, marks the API key for the ZeroSSL issuer as required.

Documentation for the fields previously denoted to be inherited by `...` should still be provided, but I am not familiar with those.

### DIFF
--- a/src/docs/markdown/caddyfile/directives/tls.md
+++ b/src/docs/markdown/caddyfile/directives/tls.md
@@ -359,22 +359,21 @@ Obtains certificates using the ACME protocol. Note that `acme` is a default issu
 
 	- **any_common_name** <span id="any_common_name"/> is a list of one or more common names; Caddy will choose the first chain that has an issuer that matches with at least one of the specified common names.
 
+##### ZeroSSL
+
+Caddy will implicitly use [ZeroSSL's ACME endpoint](https://zerossl.com/documentation/acme/) (and generate EAB credentials) if you specify the the [`email` global option](/docs/caddyfile/options#email).
+
+To provide your own EAB credentials for ZeroSSL, specify the `dir` and `eab` options.
 
 #### zerossl
 
-Obtains certificates using the ACME protocol, specifically with ZeroSSL. Note that `zerossl` is a default issuer, so configuring it explicitly is usually unnecessary.
+Obtains certificates using the ZeroSSL API.
 
 ```caddy-d
-... zerossl [<api_key>] {
+... zerossl <api_key> {
 	...
 }
 ```
-
-The syntax for `zerossl` is exactly the same as for [`acme`](#acme), except that its name is `zerossl` and it can optionally take your ZeroSSL API key.
-
-Its functionality is also the same, except that it will use ZeroSSL's directory by default and it can automatically negotiate EAB credentials (whereas with the `acme` issuer, you have to manually provide EAB credentials and set the directory endpoint).
-
-When explicitly configuring `zerossl`, configuring an `email` is required so that your certificates can appear in your ZeroSSL dashboard.
 
 #### internal
 
--- a/src/docs/markdown/caddyfile/directives/tls.md
+++ b/src/docs/markdown/caddyfile/directives/tls.md
@@ -359,15 +359,17 @@ Obtains certificates using the ACME protocol. Note that `acme` is a default issu
 
 	- **any_common_name** <span id="any_common_name"/> is a list of one or more common names; Caddy will choose the first chain that has an issuer that matches with at least one of the specified common names.
 
-##### ZeroSSL
+<aside class="tip">
 
-Caddy will implicitly use [ZeroSSL's ACME endpoint](https://zerossl.com/documentation/acme/) (and generate EAB credentials) if you specify the the [`email` global option](/docs/caddyfile/options#email).
+The `acme` issuer module will implicitly use [ZeroSSL's ACME endpoint](https://zerossl.com/documentation/acme/) (and generate EAB credentials) if you specify the the [`email` global option](/docs/caddyfile/options#email).
 
 To provide your own EAB credentials for ZeroSSL, specify the `dir` and `eab` options.
 
+</aside>
+
 #### zerossl
 
-Obtains certificates using the ZeroSSL API.
+Obtains certificates using the [ZeroSSL API](https://zerossl.com/documentation/api/).
 
 ```caddy-d
 ... zerossl <api_key> {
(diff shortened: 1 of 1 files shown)


mholt (approved):
Oh wow, thanks. Not sure how I forgot to update this!

src/docs/markdown/caddyfile/directives/tls.md
```diff
@@ -359,23 +359,22 @@ Obtains certificates using the ACME protocol. Note that `acme` is a default issu
 
 	- **any_common_name** <span id="any_common_name"/> is a list of one or more common names; Caddy will choose the first chain that has an issuer that matches with at least one of the specified common names.
 
+##### ZeroSSL
```
mholt: We can remove this section and just move the two paragraphs to under the `zerossl` header below.

kekalainen: The rationale for the seemingly redundant heading was grouping under the ACME issuer module to which those notes pertain to. Aren't they technically separate issuer modules for the same issuer?

francislavoie: Might make sense to mention ZeroSSL can be enabled by using the `acme` issuer (and putting the ZeroSSL URL in that section too) and leave the `zerossl` issuer as explaining it's for the API and not ACME, pointing back to `acme` for that. Maybe.

kekalainen: From a recently uninitiated user's perspective, it's inherently a bit confusing either way but letting the implementation dictate grouping (as drafted) seems less misleading to me. Explicitly explaining the difference could make sense indeed, though I tried to keep it terse overall thus implied it with the grouping instead.

src/docs/markdown/caddyfile/directives/tls.md
```diff
@@ -359,23 +359,22 @@ Obtains certificates using the ACME protocol. Note that `acme` is a default issu
...
 
 #### zerossl
 
-Obtains certificates using the ACME protocol, specifically with ZeroSSL. Note that `zerossl` is a default issuer, so configuring it explicitly is usually unnecessary.
+Obtains certificates using the ZeroSSL API.
 
 ```caddy-d
-... zerossl [<api_key>] {
+... zerossl <api_key> {
 	...
```
kekalainen: The omitted fields here should be documented.

mholt: Good point, here they are from the Godoc:

```go
//	... zerossl <api_key> {
//		    validity_days <days>
//		    alt_http_port <port>
//		    dns <provider_name> ...
//		    propagation_delay <duration>
//		    propagation_timeout <duration>
//		    resolvers <list...>
//		    dns_ttl <duration>
//	}
```

src/docs/markdown/caddyfile/directives/tls.md
mholt: The ACME endpoint and ZeroSSL API are two different issuers, technically. Let's not introduce any new sections and just keep the ZeroSSL description under the existing `zerossl` header.

src/docs/markdown/caddyfile/directives/tls.md
kekalainen: Added, save for describing `validity_days`.

src/docs/markdown/caddyfile/directives/tls.md
kekalainen: I removed the subsection but retained the ACME-specific configuration tips under the `acme` issuer module.

src/docs/markdown/caddyfile/directives/tls.md
mholt: Thanks. Maybe we should refer the users to the [ZeroSSL API docs](https://zerossl.com/documentation/api/create-certificate/) anyway. But for that field it's basically either 365 or 90.

src/docs/markdown/caddyfile/directives/tls.md
```diff
@@ -359,22 +359,31 @@ Obtains certificates using the ACME protocol. Note that `acme` is a default issu
 
 	- **any_common_name** <span id="any_common_name"/> is a list of one or more common names; Caddy will choose the first chain that has an issuer that matches with at least one of the specified common names.
 
+<aside class="tip">
+
+The `acme` issuer module will implicitly use [ZeroSSL's ACME endpoint](https://zerossl.com/documentation/acme/) (and generate EAB credentials) if you specify the the [`email` global option](/docs/caddyfile/options#email).
```
mholt, line 364: So, this is only true if it's unspecified in the config. If you specify an issuer in the config, that overrides the defaults. Since this is docs describing how to override the defaults, maybe it's not the place to describe the default behavior which might be confusing.

kekalainen: I see. Removing the paragraph altogether would muddy the difference between the `acme` and `zerossl` issuer modules, though. Perhaps prefixing it with "When explicitly unconfigured" would do?

mholt: What if instead, we add a sentence to the `zerossl` section that clarifies: "The ZeroSSL API is distinct from its ACME endpoint." or something like that.

kekalainen: While that'd cover the distinction more directly, the `acme` section still claims the module is "using Let's Encrypt" by default. For consistency either all or none of the possible implicit/default configuration should be mentioned IMO.

Especially since the [release notes](https://github.com/caddyserver/caddy/releases/tag/v2.8.0) claim that when the `email` global is configured,

> you don't have to make any changes and you'll still get Let's Encrypt and ZeroSSL automatically as defaults.

which, turns out, is arguably not true if `acme` is explicitly configured. That's not apparent (to me anyway) when said configuration is not providing a `directory_url` as an argument nor field value. (For context, I use the `acme` issuer directive to override the DNS resolvers used for challenges (since those aren't globally configurable AFAIK) and had the false intuition it'd leave the dirs intact.)

mholt: >  the acme section still claims the module is "using Let's Encrypt" by default. For consistency either all or none of the possible implicit/default configuration should be mentioned IMO.

Fair point, maybe that section needs to be tweaked to say "(using Let's Encrypt, and if an email is provided, ZeroSSL too)" or similar.

> which, turns out, is arguably not true if acme is explicitly configured. That's not apparent (to me anyway) when said configuration is not providing a directory_url as an argument nor field value.

Perhaps verbiage clarifying that specifying any issuers wipes out implicit defaults.

francislavoie: What's the status on this at this point?

mholt: I think we're still waiting for a few tweaks to be made; but if not, I might try to wrap this up myself at some point. Although, I don't think I have push rights to the branch.

mholt:
Yeah I can't push to the branch, so, closing due to inactivity.

PR #419, opened 2024-09-29
Finish documenting Caddy 2.8.0 features

francislavoie:
This is... very late 🙈 I've been unmotivated to get to this, but I'm finally getting around to completing all this. I was only about halfway done when 2.8.0 first released. This is the second half of the changes in https://github.com/caddyserver/caddy/compare/v2.7.6...v2.8.0

FYI @dunglas @armadi1809 I wouldn't mind a quick look at the bits you guys worked on just in case I missed anything. Wouldn't mind more example ideas if you have any.

Notably @mholt I added a new page for Response Matchers which is now in the left nav, this is extracted from the Reverse Proxy and Encode pages which duplicated this information. Since we were also adding it to Intercept, it was time to split it off on its own page. I also moved Global Options to be above Directives in the nav to match the order they appear in a Caddyfile and in the Concepts page. I think that makes more sense.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,3 +1,5 @@
+root = true
+
 [*]
 end_of_line = lf
 insert_final_newline = true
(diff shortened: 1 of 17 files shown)


francislavoie:
I'll probably just merge this now to get it out, but feedback is still very much welcome, and I'll do followup commits as needed.

mholt:
Thank you Francis. Sorry I'm behind. Weekends :sweat_smile:

dunglas:
Thanks @francislavoie! The response intercepting part LGTM.

PR #425, opened 2024-10-10
expand the `listener_wrapper` docs

### LINKED ISSUE #398: Missing additional `proxy_protocol` docs
there are additional configuration options (i.e. `deny`, `fallback_policy`) that were not added to the documentation

_Originally posted by @mohammed90 in https://github.com/caddyserver/caddy/issues/6390#issuecomment-2162263308_
            
Creating this as reminder for myself to add the docs.

mohammed90:
With this PR, it looks like this:

image

Closes #398

### DIFF
--- a/src/docs/markdown/caddyfile/options.md
+++ b/src/docs/markdown/caddyfile/options.md
@@ -787,21 +787,6 @@ http:// {
 
 Allows configuring [listener wrappers](/docs/json/apps/http/servers/listener_wrappers/), which can modify the behaviour of the socket listener. They are applied in the given order.
 
-There is a special no-op [`tls`](/docs/json/apps/http/servers/listener_wrappers/tls/) listener wrapper provided as a standard module which marks where TLS should be handled in the chain of listener wrappers. It should only be used if another listener wrapper must be placed in front of the TLS handshake. This _does not_ enable TLS for a server; e.g. if this is used on your `:80` HTTP server, it will still act as a no-op.
-
-The included [`http_redirect`](/docs/json/apps/http/servers/listener_wrappers/http_redirect/) listener wrapper can look at the first few bytes of an incoming request to determine if it's likely HTTP (instead of TLS), and trigger an HTTP-to-HTTPS redirect on the same port but using the `https://` scheme. This is most useful when serving HTTPS on a non-standard port (other than `443`), since browsers will try HTTP unless the scheme is specified. Don't use this on an HTTP server. It must be placed _before_ the `tls` listener wrapper. For example:
-
-```caddy
-{
-	servers {
-		listener_wrappers {
-			http_redirect
-			tls
-		}
-	}
-}
-```
-
 ###### `tls`
 
 The `tls` listener wrapper is a no-op listener wrapper that marks where the TLS listener should be in a chain of listener wrappers. It should only be used if another listener wrapper must be placed in front of the TLS handshake.
--- a/src/docs/markdown/caddyfile/options.md
+++ b/src/docs/markdown/caddyfile/options.md
@@ -802,6 +802,8 @@ The included [`http_redirect`](/docs/json/apps/http/servers/listener_wrappers/ht
 }
 ```
 
+###### `proxy_protocol`
+
 Also included is the [`proxy_protocol`](/docs/json/apps/http/servers/listener_wrappers/proxy_protocol/) listener wrapper (prior to v2.7.0 it was only available via a plugin), which enables [PROXY protocol](https://github.com/haproxy/haproxy/blob/master/doc/proxy-protocol.txt) parsing (popularized by HAProxy). This must be used _before_ the `tls` listener wrapper since it parses plaintext data at the start of the connection:
 
 ```caddy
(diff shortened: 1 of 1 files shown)


src/docs/markdown/caddyfile/options.md
```diff
@@ -811,13 +813,29 @@ Also included is the [`proxy_protocol`](/docs/json/apps/http/servers/listener_wr
...
+- **allow** is a list of CIDR ranges of trusted sources to receive PROXY headers. Unix sockets are trusted by default and not part of this option.
+
+- **deny** is a list of CIDR ranges of trusted sources to reject PROXY headers from.
+
+- **fallback_policy** is the action to take if the PROXY header comes from an address that not in either list of allow/deny. The default fallback policy is `IGNORE`. Accepted values of `fallback_policy` are:
+	- IGNORE: address from PROXY header, but accept connection
+	- USE: address from PROXY header
+	- REJECT: connection when PROXY header is sent
+	- REQUIRE: connection to send PROXY header, reject if not present
+	- SKIP: accepts a connection without requiring the PROXY header.
```
```suggestion
	- `IGNORE`: address from PROXY header, but accept connection
	- `USE`: address from PROXY header
	- `REJECT`: connection when PROXY header is sent
	- `REQUIRE`: connection to send PROXY header, reject if not present
	- `SKIP`: accepts a connection without requiring the PROXY header.
```

src/docs/markdown/caddyfile/options.md
```diff
@@ -802,6 +802,8 @@ The included [`http_redirect`](/docs/json/apps/http/servers/listener_wrappers/ht
 }
 ```
 
+###### `proxy_protocol`
```
francislavoie, line 809: For consistency I think we should add a `http_redirect` heading above, and maybe adjust it to not phrase it like "also" and "included" etc

mohammed90: Done

francislavoie: Thanks!

https://github.com/caddyserver/website/pull/425/files#diff-b3c7d3ecc614554a34ea7d0bf347589a8cc02570b4821508013cf68e9109d304R788-R792 is a bit redundant now tho (lines 788-792)

src/docs/markdown/caddyfile/options.md
```diff
@@ -787,22 +787,17 @@ http:// {
...
-		listener_wrappers {
-			http_redirect
-			tls
-		}
-	}
-}
-```
+###### `http_redirect`
+
+The [`http_redirect`](/docs/json/apps/http/servers/listener_wrappers/http_redirect/) provides HTTP->HTTPS redirects for connections that come on the TLS port as an HTTP request, by detecting using the first few bytes that it's not a TLS handshake, but instead an HTTP request. This is most useful when serving HTTPS on a non-standard port (other than `443`), since browsers will try HTTP unless the scheme is specified. It must be placed _before_ the `tls` listener wrapper.
```
francislavoie: I'm not able to make this a github suggestion because it contains a code block :rofl: but I think we should re-add the `http_redirect` example:

```caddy
{
	servers {
		listener_wrappers {
			http_redirect
			tls
		}
	}
}
```

src/docs/markdown/caddyfile/options.md
```diff
@@ -811,13 +806,29 @@ Also included is the [`proxy_protocol`](/docs/json/apps/http/servers/listener_wr
 			proxy_protocol {
 				timeout 2s
 				allow 192.168.86.1/24 192.168.86.1/24
+				deny 10.0.0.0/8
```
francislavoie, line 845: This reads like an example. I think it should be changed to the "Syntax" form (like `timeout <duration>` etc) and then move this example below the options list.

src/docs/markdown/caddyfile/options.md
```diff
@@ -811,13 +806,29 @@ Also included is the [`proxy_protocol`](/docs/json/apps/http/servers/listener_wr
...
 }
 ```
 
+- **timeout** specifies the maximum duration to wait for the PROXY header. Defaults to `5s`.
+
+- **allow** is a list of CIDR ranges of trusted sources to receive PROXY headers. Unix sockets are trusted by default and not part of this option.
+
+- **deny** is a list of CIDR ranges of trusted sources to reject PROXY headers from.
+
+- **fallback_policy** is the action to take if the PROXY header comes from an address that not in either list of allow/deny. The default fallback policy is `IGNORE`. Accepted values of `fallback_policy` are:
```
francislavoie: Could mention "case insensitive"

src/docs/markdown/caddyfile/options.md
mohammed90: I almost added it, then decided it not to in the last second 😂  I'll re-add it

src/docs/markdown/caddyfile/options.md
mohammed90: I was afraid of encouraging the users, but fine if we can commit to it.

francislavoie: Well the example had it as lowercase `fallback_policy reject` :joy: if you prefer just making that consistent instead of mentioning insensitive, that's fine too

francislavoie: I think if we document it as lowercase it might be more Caddyfile-like, we don't usually use uppercase for constants/options in Caddyfile.

src/docs/markdown/caddyfile/options.md
mohammed90: Better?

src/docs/markdown/caddyfile/options.md
mohammed90: Done

src/docs/markdown/caddyfile/options.md
```diff
@@ -787,11 +787,15 @@ http:// {
...
+The `tls` listener wrapper is a no-op listener wrapper that marks where the TLS listener should be in a chain of listener wrappers. It should only be used if another listener wrapper must be placed in front of the TLS handshake.
 
-```caddy
+###### `http_redirect`
+
+The [`http_redirect`](/docs/json/apps/http/servers/listener_wrappers/http_redirect/) provides HTTP->HTTPS redirects for connections that come on the TLS port as an HTTP request, by detecting using the first few bytes that it's not a TLS handshake, but instead an HTTP request. This is most useful when serving HTTPS on a non-standard port (other than `443`), since browsers will try HTTP unless the scheme is specified. It must be placed _before_ the `tls` listener wrapper. Here's an example:
+
+```
```
```suggestion
```caddy
```

src/docs/markdown/caddyfile/options.md
francislavoie: Did a commit to do what I mean :sweat_smile:

PR #428, opened 2024-10-28
Fix forward_auth.md example

teodorescuserban:
`request_header` doesn't seem to accept multiple headers at a time.


mholt (approved):
Ah thanks, this is true; though probably an improvement that could be made for the syntax as well, to work like `header` (where you can specify a block).

teodorescuserban:
Indeed!

PR #430, opened 2024-12-12
tls: Replace usage of deprecated field `trusted_ca_cert_file`

Pyker:
I used this example and, after running `caddy validate`, noticed the following warning, so I figured I might as well just submit a PR updating it:

```
WARN    The 'trusted_ca_cert_file' field is deprecated. Use the 'trust_pool' field instead.
```

Additionally, I couldn't find any documentation for the `trusted_ca_cert_file` field, so I'm not sure if this replacement is correct or not.

### DIFF
--- a/src/docs/markdown/caddyfile/directives/tls.md
+++ b/src/docs/markdown/caddyfile/directives/tls.md
@@ -514,7 +514,6 @@ Enable TLS Client Authentication and require clients to present a valid certific
 example.com {
 	tls {
 		client_auth {
-			mode       require_and_verify
 			trust_pool file ../caddy.ca.cer ../root.ca.cer
 		}
 	}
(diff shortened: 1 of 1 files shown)


Pyker:
It does show up in the search, but it's nowhere in the actual page (<https://caddyserver.com/docs/caddyfile/directives/tls#syntax>), besides the example I just edited: 
image

Pyker:
From what I understand, that was removed in https://github.com/caddyserver/website/commit/ca9ce7b2b527bcd2d7031cac29c5cfef2e865a62

mohammed90:
Thanks! The search function is via 3rd party that, as I understand it, scrapes the pages to be indexed. They probably still have it cached.

mohammed90 (approved):
LGTM!

mholt:
(Thanks; yeah, our site search is less than excellent... but it's also free, so I can't complain. I do hardly use it though and mainly stick to Google.)

Pyker:
> (Thanks; yeah, our site search is less than excellent... but it's also free, so I can't complain. I do hardly use it though and mainly stick to Google.)

(It's okay, I figured I should probably point it out in case it was something under your control)

PR #432, opened 2024-12-15
listing all possible log levels

yene:


mholt (commented):
Is this really necessary since the only levels emitted are INFO and ERROR, which it says on the line below?

francislavoie:
This page has dual purpose of the `log` directive and `log` global option, so it does make sense to document, but it should be clear which levels are useful when. But yes, only `INFO` and `ERROR` for access logs right now, other levels for global logs.

steffenbusch:
FYI: on global log option, it has been already added https://github.com/caddyserver/website/commit/1cc9fecaccd57fec3644cd712334135d9d738ac5

(I was once asking this myself and opened an Issue and that commit fixed it)

francislavoie:
Right, in that case this isn't necessary. Thanks!

PR #433, opened 2024-12-30
Update compose.yml to match Docker Hub

eosti:
Changes the location of the Caddyfile to within a `conf` directory instead of being directly mounted, as suggested by https://github.com/docker-library/docs/pull/2483.

This issue was found in https://github.com/caddyserver/caddy/issues/5735 and was fixed in the Docker Hub documentation, but the Caddy documentation still referred to the old method.

### DIFF
--- a/src/docs/markdown/running.md
+++ b/src/docs/markdown/running.md
@@ -293,7 +293,7 @@ services:
       - "443:443"
       - "443:443/udp"
     volumes:
-      - ./Caddyfile:/etc/caddy/Caddyfile
+      - ./conf:/etc/caddy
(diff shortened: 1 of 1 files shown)


francislavoie:
Thanks! I'll take a closer look "next year" 😅 in a couple days when I'm back home

mholt (approved):
I think this is probably good, yeah?

francislavoie:
Ah yeah :see_no_evil: sorry I never got back around to this

mholt:
Haha no worries, I hadn't either!

PR #434, opened 2025-01-01
sort submodules lists

mohammed90:
This has been bothering me for some time and finally bit the bullet to fix it. The submodules were listed without specific order, making looking for a specific module in the list hard. Now they're sorted: standard modules first, then alphabetically.

**Before:**

image

**After:**

image


mholt (approved):
Thanks! Nice improvement.

mholt (approved):

PR #440, opened 2025-01-06
Update import.md

paulovieira:
See details here: https://github.com/caddyserver/caddy/pull/6130#issuecomment-2569999916


mholt (approved):
Thanks!

src/docs/markdown/caddyfile/directives/import.md
```diff
@@ -94,34 +94,62 @@ example.com {
...
+```
+
+Import a snippet which provides extendable options for a reverse proxy:
 
 ```caddy
 (extendable-proxy) {
 	reverse_proxy {
-		to {block.proxy_target}
+		{block.proxy_target}
 		{block.proxy_options}
```
francislavoie, line 129: These should be `blocks` too, right?

paulovieira: > These should be `blocks` too, right?

Right, I missed those, thanks. A new PR had been created.

PR #441, opened 2025-01-07
Update import.md

paulovieira:
Adjustments for the `import` directive:

 - replace `{block...}` with `{block}`, because in the initial "Syntax" code block we have `{block}`
 - in the example, correct `{block.proxy_target}` and `{block.proxy_options}`


mholt (approved):
I was wondering about that. :sweat_smile:  Thanks

PR #442, opened 2025-01-08
docs: document more of `header` directive behavior

lilnasy:
- Added syntax, usage, and examples for the `match` directive added in 2.9.1 https://github.com/caddyserver/caddy/pull/6765
- Clarified in a few places that the headers are being added/removed/replaced in the responses, not the requests.

### DIFF
--- a/src/docs/markdown/caddyfile/directives/encode.md
+++ b/src/docs/markdown/caddyfile/directives/encode.md
@@ -31,7 +31,7 @@ encode [<matcher>] <formats...> {
 
 	match {
 		status <code...>
-		header <field> <value>
+		header <field> [<value>]
 	}
 }
 ```
--- a/src/docs/markdown/caddyfile/directives/encode.md
+++ b/src/docs/markdown/caddyfile/directives/encode.md
@@ -29,10 +29,7 @@ encode [<matcher>] <formats...> {
 	
 	minimum_length <length>
 
-	match {
-		status <code...>
-		header <field> [<value>]
-	}
+	match <inline_response_matcher>
 }
 ```
 
--- a/src/docs/markdown/caddyfile/directives/encode.md
+++ b/src/docs/markdown/caddyfile/directives/encode.md
@@ -31,7 +31,7 @@ encode [<matcher>] <formats...> {
 
 	match {
 		status <code...>
-		header <field> [<value>]
+		header <field> <value>
 	}
 }
 ```
--- a/src/docs/markdown/caddyfile/directives/header.md
+++ b/src/docs/markdown/caddyfile/directives/header.md
@@ -130,9 +130,7 @@ reverse_proxy upstream:443
 Mark all successful responses to GET requests as cacheable for upto an hour:
 
 ```caddy-d
-@GET {
-	method GET
-}
+@GET method GET
 header @GET Cache-Control "max-age=3600" {
 	match status 2xx
 }
--- a/src/docs/markdown/caddyfile/directives/header.md
+++ b/src/docs/markdown/caddyfile/directives/header.md
@@ -37,6 +37,15 @@ header [<matcher>] [[+|-|?|>]<field> [<value>|<find>] [<replace>]] {
 	?<field> <value>
 
 	[defer]
+
+	# a match condition
+	match [status <code...> | header <field> <value>]
+
+	# or a match block
+	match {
+		status <code...>
+		header <field> <value>
+	}
 }
 ```
 
@@ -44,13 +53,13 @@ header [<matcher>] [[+|-|?|>]<field> [<value>|<find>] [<replace>]] {
 
   With no prefix, the field is set (overwritten).
 
-  Prefix with `+` to add the field instead of overwriting (setting) the field if it already exists; header fields can appear more than once in a request.
+  Prefix with `+` to add the field instead of overwriting (setting) the field if it already exists; header fields can appear more than once in a response.
 
   Prefix with `-` to delete the field. The field may use prefix or suffix `*` wildcards to delete all matching fields.
 
   Prefix with `?` to set a default value for the field. The field is only written if it doesn't yet exist.
 
-  Prefix with `>` to set the field, and enable `defer`, as a shortcut.
+  Prefix with `>` to replace the contents of an existing header value using a regular expression. Implies `defer`, see below.
 
 - **&lt;value&gt;** is the header field value, when adding or setting a field.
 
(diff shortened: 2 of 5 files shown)


src/docs/markdown/caddyfile/directives/encode.md
```diff
@@ -31,7 +31,7 @@ encode [<matcher>] <formats...> {
 
 	match {
 		status <code...>
-		header <field> [<value>]
+		header <field> <value>
```
lilnasy: I might be reading this wrong, but this seemed to suggest that the header value is optional. This change to emphasize that both the header name and the value are required.

francislavoie: It's technically optional because you can do `-Field` to delete a header, in which case no value is passed.

lilnasy: Are you talking about the header directive?

francislavoie: Oh, yeah sorry. I didn't click through to the files changed so I didn't notice it was the response matcher in `encode`. That said, `!` as a prefix is allowed, meaning "this header doesn't exist in the response" and in that case you omit the value. See https://caddyserver.com/docs/caddyfile/response-matchers#header

src/docs/markdown/caddyfile/directives/header.md
```diff
@@ -37,28 +37,43 @@ header [<matcher>] [[+|-|?|>]<field> [<value>|<find>] [<replace>]] {
 	?<field> <value>
 
 	[defer]
+
+	# a match condition
+	match [status <code...> | header <field> <value>]
```
francislavoie: This one's a bit strange cause it's not optional to provide a matcher, so the `[]` are sorta misplaced.

I think we should probably replace these to just `<response_matcher...>` instead, similar to how we have `<matcher>` in places where a request matcher can go. We can add a jquery blob at the top to transform those into links to https://caddyserver.com/docs/caddyfile/response-matchers

src/docs/markdown/caddyfile/directives/header.md
```diff
@@ -37,28 +37,43 @@ header [<matcher>] [[+|-|?|>]<field> [<value>|<find>] [<replace>]] {
...
 
-  Prefix with `+` to add the field instead of overwriting (setting) the field if it already exists; header fields can appear more than once in a request.
+  Prefix with `+` to add the field instead of overwriting (setting) the field if it already exists; header fields can appear more than once in a response.
 
   Prefix with `-` to delete the field. The field may use prefix or suffix `*` wildcards to delete all matching fields.
 
   Prefix with `?` to set a default value for the field. The field is only written if it doesn't yet exist.
 
-  Prefix with `>` to set the field, and enable `defer`, as a shortcut.
+  Prefix with `>` to replace the contents of an existing header value using a regular expression. Implies `defer`, see below.
```
francislavoie: Huh? That's not true. `>` is literally only to enable `defer`. It does nothing with regexp.

The regexp replace happens if there's 3 arguments, where the first is the field, the second is a search and third is the replace.

Take a look again at the example patterns above, i.e. `# Set with defer` `><field> <value>` and `# Replace with defer` `><field> <find> <replace>` which are counterparts to their "without defer" versions when `>` is omitted.

src/docs/markdown/caddyfile/directives/header.md
```diff
@@ -37,28 +37,43 @@ header [<matcher>] [[+|-|?|>]<field> [<value>|<find>] [<replace>]] {
...
+	}
 }
 ```
 
 - **&lt;field&gt;** is the name of the header field.
 
   With no prefix, the field is set (overwritten).
 
-  Prefix with `+` to add the field instead of overwriting (setting) the field if it already exists; header fields can appear more than once in a request.
+  Prefix with `+` to add the field instead of overwriting (setting) the field if it already exists; header fields can appear more than once in a response.
```
francislavoie, line 49: :+1: right

src/docs/markdown/caddyfile/directives/header.md
```diff
@@ -37,28 +37,43 @@ header [<matcher>] [[+|-|?|>]<field> [<value>|<find>] [<replace>]] {
...
 
 - **&lt;find&gt;** is the substring or regular expression to search for.
 
 - **&lt;replace&gt;** is the replacement value; required if performing a search-and-replace. Use `$1` or `$2` and so on to reference capture groups from the search pattern. If the replacement value is `""`, then the matching text is removed from the value. See the [Go documentation](https://golang.org/pkg/regexp/#Regexp.Expand) for details.
 
-- **defer** will force the header operations to be deferred until the response is being written out to the client. This is automatically enabled if any of the header fields are being deleted with `-`, when setting a default value with `?`, or when having used the `>` prefix.
+- **defer** defers the execution of header operations until the response is being sent to the client. This option is automatically enabled under the following conditions:
+	- When any header fields are deleted using `-`.
+	- When setting a default value with `?`.
+	- When using the `>` prefix to replace within the header value.
```
francislavoie: Same here, `>` has no implication of replacement. I'd write here `When the `>` prefix is used as a shortcut.`

src/docs/markdown/caddyfile/directives/encode.md
lilnasy: Thanks, didn't know about the `!` syntax. I'll revert.

src/docs/markdown/caddyfile/directives/header.md
lilnasy: Thanks, I had the wrong idea here.

src/docs/markdown/caddyfile/directives/header.md
lilnasy: I think that's a good idea. There are two ways in which response matchers can be used. I documented the difference.

<image src="https://github.com/user-attachments/assets/ec42795c-4299-4830-83ae-28085f7e24d0" width="400"/>

`encode` and `headers` allow inline matching, `intercept` requires creating named response matchers.

mholt (approved):
Thanks for iterating on this! @francislavoie Does this LGTY?

src/docs/markdown/caddyfile/directives/header.md
```diff
@@ -119,10 +127,61 @@ header ?Cache-Control "max-age=3600"
 reverse_proxy upstream:443
 ```
 
+Mark all successful responses to GET requests as cacheable for upto an hour:
+
+```caddy-d
+@GET {
+	method GET
+}
```
francislavoie:
```suggestion
@GET method GET
```

src/docs/markdown/caddyfile/response-matchers.md
```diff
@@ -35,14 +35,39 @@ These typically only appear as config inside of certain other directives, to mak
...
+	status <code...>
+	header <field> [<value>]
+}
+```
+```caddy-d
+status <code...>
+```
+```caddy-d
+header <field> [<value>]
+```
```
francislavoie:
````suggestion
```caddy-d
... {
	status <code...>
	header <field> [<value>]
}
```
```caddy-d
... status <code...>
```
```caddy-d
... header <field> [<value>]
```
````

francislavoie (approved):

PR #455, opened 2025-02-28
Add metrics per_host documentation to caddy website.

RaspberryProgramming:
I saw that per_host was an option that I didn't see in docs. I thought it'd be useful to add this so others don't have a hard time discovering it exists.

### DIFF
--- a/src/docs/markdown/caddyfile/options.md
+++ b/src/docs/markdown/caddyfile/options.md
@@ -971,7 +971,7 @@ Enables Prometheus metrics collection; necessary before scraping metrics. Note t
 }
 ```
 
-You can add or remove the per_host option to separate metrics for each host rather than for the entire caddy server.
+You can add the `per_host` option to label metrics with the host name of the metric.
 
 ```caddy
 {
--- a/src/docs/markdown/caddyfile/options.md
+++ b/src/docs/markdown/caddyfile/options.md
@@ -76,7 +76,9 @@ Possible options are (click on each option to jump to its documentation):
 	}
 	grace_period   <duration>
 	shutdown_delay <duration>
-	metrics
+	metrics {
+        per_host
+    }
 
 	# TLS Options
 	auto_https off|disable_redirects|ignore_loaded_certs|disable_certs
--- a/src/docs/markdown/caddyfile/options.md
+++ b/src/docs/markdown/caddyfile/options.md
@@ -969,6 +969,15 @@ Enables Prometheus metrics collection; necessary before scraping metrics. Note t
 }
 ```
 
+You can add or remove the per_host option to separate metrics for each host rather than for the entire caddy server.
+
+```caddy
+{
+	metrics {
+        per_host
+    }
+}
+```
 
 ##### `trace`
 
--- a/src/docs/markdown/caddyfile/options.md
+++ b/src/docs/markdown/caddyfile/options.md
@@ -77,8 +77,8 @@ Possible options are (click on each option to jump to its documentation):
 	grace_period   <duration>
 	shutdown_delay <duration>
 	metrics {
-        per_host
-    }
+		per_host
+	}
 
 	# TLS Options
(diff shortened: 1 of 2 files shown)


mholt:
Thanks -- IIRC there's some caveats to this though. Should we add those notes in the docs? (I didn't work on this feature so I'm not familiar, but I've requested reviews from others who might be.)

src/docs/markdown/metrics.md
```diff
@@ -25,6 +25,18 @@ If using a Caddyfile, enable metrics [in global options](/docs/caddyfile/options
...
 
+To add per host metrics you can insert the per_host option. Host specific metrics will now have a Host tag.
+
+```caddy
+{
+    servers {
+        metrics {
+            per_host
+        }
+    }
```
mohammed90: The metrics config is now moved outside of the `servers` option.

mohammed90 (changes requested):
There's an entry for `metrics` at the top of the file where all the global options are. Can you please add the block with `per_host` inside it? Thanks!

src/docs/markdown/metrics.md
mholt: D'oh, I should have tagged you Mohammed :man_facepalming:

RaspberryProgramming: Do you want me to update the code block above the one I added also?

RaspberryProgramming: This has been fixed. I've also added the requested change.

RaspberryProgramming:
@mholt 
> Thanks -- IIRC there's some caveats to this though. Should we add those notes in the docs? (I didn't work on this feature so I'm not familiar, but I've requested reviews from others who might be.)

I believe the only caveat I can think of is what was described already in the documentation.
On line 966: "Note that metrics reduce performance on really busy servers. (Our community is working on improving this. Please get involved!)"

In testing the metrics basically get split up per each host in your caddy file instead of on a single metric. I could possibly include a sample of what the metrics look like with or without the per_host option but it might end up making this document messy.

src/docs/markdown/caddyfile/options.md
```diff
@@ -969,6 +971,15 @@ Enables Prometheus metrics collection; necessary before scraping metrics. Note t
 }
 ```
 
+You can add or remove the per_host option to separate metrics for each host rather than for the entire caddy server.
```
mohammed90: This is not accurate. When the `per_host` option is added, it adds a label of the hostname for which the metrics is recorded. This way you can see metrics for a particular hostname.

```suggestion
You can add the `per_host` option to label metrics with the host name of the metric.
```

src/docs/markdown/metrics.md
```diff
@@ -17,14 +17,22 @@ If using a Caddyfile, enable metrics [in global options](/docs/caddyfile/options
...
-    servers {
-        metrics
-    }
+	metrics
 }
 ```
 
 If using JSON, add `"metrics": {}` to your [`apps > http > servers` configuration](/docs/json/apps/http/servers/).
 
+To add per host metrics you can insert the per_host option. Host specific metrics will now have a Host tag.
```
mohammed90:
```suggestion
To add per-host metrics you can insert the `per_host` option. Host specific metrics will now have a Host tag.
```

mohammed90 (approved):
Looks good to me now 💯 I'll leave it for 1 more pair of eyes to check

src/docs/markdown/caddyfile/options.md
```diff
@@ -969,6 +971,15 @@ Enables Prometheus metrics collection; necessary before scraping metrics. Note t
 }
 ```
 
+You can add the `per_host` option to label metrics with the host name of the metric.
+
+```caddy
+{
+	metrics {
+        per_host
+    }
```
francislavoie: spaces -> tabs
```suggestion
		per_host
	}
```

francislavoie (approved):

PR #463, opened 2025-03-29
Prevent Horizontal Overflow on Homepage when using a Phone

ff2400t:
This fixes a Horizontal Overflow that appears on Mobile. Here is an Example of that
Screenshot_20250329-130012


mholt (approved):
Thanks -- I don't think this will _actually_ solve the problem, even if it solves it for phones that are as narrow as 300px, but any narrower and I imagine it'll still happen. But I guess we can do this for now. :)

PR #467, opened 2025-04-23
Strip headers when proxying errors to external service

JEHoctor:
I think this example error handling system from the docs might send potentially sensitive headers to an untrusted third party. This is my attempt to recommend stripping those headers in the request to http.cat. Thank you!

### DIFF
--- a/src/docs/markdown/caddyfile/directives/handle_errors.md
+++ b/src/docs/markdown/caddyfile/directives/handle_errors.md
@@ -83,6 +83,7 @@ Reverse proxy to a professional server that is highly qualified for handling HTT
 handle_errors {
 	rewrite * /{err.status_code}
 	reverse_proxy https://http.cat {
+		header_up -*
 		header_up Host {upstream_hostport}
 		replace_status {err.status_code}
 	}


francislavoie:
No, this isn't a good example. This would break 99% of legitimate web requests.

JEHoctor:
@francislavoie

Based on the docs [here](https://caddyserver.com/docs/caddyfile/directives/reverse_proxy#defaults), I believe that the example as it exists in the docs today sends all headers from the original request to http.cat, with the only exception being the Host header. So if the request contains, for example, a session cookie, that would be sent to http.cat.

I'm seeking to avoid the following undesirable scenario:

1. A user configures Caddy as a reverse proxy to a service that uses session cookies
2. The user configures this handle_errors block as found in the docs
3. A bad request includes a session cookie and triggers this handle_errors directive
4. The session cookies are sent to http.cat
5. The owners of http.cat may at some point in the past, present, or future collect this information for malicious purposes, or their systems may be compromised by someone wishing to exploit this information

I have tried this change with two different Caddy deployments, and I can verify that the 404 page from http.cat is served just the same as before. On the other hand, I have not attempted to directly observe any headers leaking to http.cat.

Please excuse my ignorance on this topic. This is my attempt to engage with an unfamiliar topic in good faith. I very much appreciate your time.

mholt:
> So if the request contains, for example, a session cookie, that would be sent to http.cat.

Why would you send a session cookie to http.cat though? Like, don't do that. There are much better ways to not do that.

JEHoctor:
@mholt, ah this is likely what I don't understand. What is the proper way to prevent that? From the client's perspective, they would be sending the cookie to the service they want to authenticate with, so I don't see how you can robustly prevent this on the client side.

Thanks again for engaging with this, I wasn't able to find a good resource on this topic. RTFM would be my preference over engaging with developers if I knew how to find TFM for this question.

mholt:
Like how to prevent passing unwanted headers? You can remove them using `header_up -`, but you wouldn't want to remove _all_ of them (unless you re-add all of the non-sensitive ones, which might be intractable).

This might be a better discussion for the forum: https://caddy.community where it will get more eyes on it.

PR #468, opened 2025-04-27
Add documentation for the `set` option in `request_body` directive

steffenbusch:
With Caddy v2.10.0, the `set` option was introduced to the `request_body` directive (see [caddyserver/caddy#5795](https://github.com/caddyserver/caddy/pull/5795)).

This PR updates the documentation to include details about the `set` option, explaining its functionality and providing an example of its usage.

### DIFF
--- a/src/docs/markdown/caddyfile/directives/request_body.md
+++ b/src/docs/markdown/caddyfile/directives/request_body.md
@@ -17,7 +17,7 @@ request_body [<matcher>] {
 
 - **max_size** is the maximum size in bytes allowed for the request body. It accepts all size values supported by [go-humanize](https://pkg.go.dev/github.com/dustin/go-humanize#pkg-constants). Reads of more bytes will return an error with HTTP status `413`.
 
-⚠️ <i>Experimental</i> <span style='white-space: pre;'> | </span> <span>v2.10.x+</span>
+⚠️ <i>Experimental</i> <span style='white-space: pre;'> | </span> <span>v2.10.0+</span>
 - **set** allows setting the request body to specific content. The content can include placeholders to dynamically insert data.
 
 ## Examples
@@ -39,7 +39,7 @@ Set the request body with a JSON structure containing a SQL query:
 example.com {
 	handle /jazz {
 		request_body {
-			set `\{"statementText":"SELECT name, genre, debut_year FROM artists WHERE genre = 'Jazz'"\}`
+			set `\{"statementText":"SELECT name, genre, debut_year FROM artists WHERE genre = 'Jazz'"}`
 		}
 
 		reverse_proxy localhost:8080 {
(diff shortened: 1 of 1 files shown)


src/docs/markdown/caddyfile/directives/request_body.md
```diff
@@ -6,17 +6,19 @@ title: request_body (Caddyfile directive)
...
 ```caddy-d
 request_body [<matcher>] {
 	max_size <value>
+	set <body_content>
 }
 ```
 
 - **max_size** is the maximum size in bytes allowed for the request body. It accepts all size values supported by [go-humanize](https://pkg.go.dev/github.com/dustin/go-humanize#pkg-constants). Reads of more bytes will return an error with HTTP status `413`.
 
+⚠️ <i>Experimental</i> <span style='white-space: pre;'> | </span> <span>v2.10.x+</span>
```
francislavoie: Can just say `.0` instead of `.x` cause we know the exact version it was added

```suggestion
⚠️ <i>Experimental</i> <span style='white-space: pre;'> | </span> <span>v2.10.0+</span>
```

src/docs/markdown/caddyfile/directives/request_body.md
```diff
@@ -30,3 +32,21 @@ example.com {
...
+		request_body {
+			set `\{"statementText":"SELECT name, genre, debut_year FROM artists WHERE genre = 'Jazz'"\}`
+		}
+
+		reverse_proxy localhost:8080 {
+			header_up Content-Type application/json
+			method POST
+			rewrite * /execute-sql
+		}
+    }
```
francislavoie: Space -> tab
```suggestion
	}
```

src/docs/markdown/caddyfile/directives/request_body.md
```diff
@@ -30,3 +32,21 @@ example.com {
...
 }
 ```
+
+Set the request body with a JSON structure containing a SQL query:
+
+```caddy
+example.com {
+	handle /jazz {
+		request_body {
+			set `\{"statementText":"SELECT name, genre, debut_year FROM artists WHERE genre = 'Jazz'"\}`
```
francislavoie: I'm pretty sure only the opening `{` needs escaping, but actually probably neither of them need it since there's spaces in this example.

steffenbusch: 👍  Just verified, only the opening `{` needs escaping. I will adjust the example.

I have tested with this config, where placeholders are actually used (but I don't want to use placeholders in the SQL example, so as not to encourage bad practices)

```
set `{"countryCode": "{country}","vatNumber": "{vatNumber}"}`
```

If, I don't escape the first `{`, I only get something like `","vatNumber": "22046007"}`.

If I would not escape the opening `{` like in an example without placeholders, like in this example:
```
 set  `{"countryCode": "XX","vatNumber": "ZZ}"}`
```
I would only get `"}`.

mholt:
Thank you @steffenbusch ! (And @francislavoie for the review)

PR #473, opened 2025-05-16
docs: query match: add missing documentation

OmarAhmadyar:
Add proper documentation for behavior implemented in [caddy/modules/caddyhttp/matchers.go](https://github.com/caddyserver/caddy/blob/master/modules/caddyhttp/matchers.go) lines 832, 848, and 859.


OmarAhmadyar:
A matching entry for expression query is also likely in need.

mholt (approved):
Thanks for the clarifications!

PR #476, opened 2025-05-29
Update install docs for newer releases of RHEL and derivatives

johankok:
The existing installation instructions for Fedora and RHEL/CentOS 8 also apply to newer versions of RHEL and its derivatives, including versions 9 and 10. This PR updates the page to clarify that compatibility.


mholt:
Thanks!

PR #480, opened 2025-06-18
Add embedded playgrounds on handle and handle_path

ProfessorLogout:
Hey @mholt and team,

as discussed I took a stab at adding [tech playground embedded playgrounds](https://tech-playground.com/docs/embedding/) as we discussed in chat.

I've added it two slightly different ways:

- `handle_path`: Embedded into one of the examples
- `handle`: Added a separate header "Playground" after the examples, building on it to make it more useful as a playground

I did not do any kind of styling so far, the element works mostly OK just with your existing styles.
If we decide to do this as a feature in the docs I would like to take some time to polish things with a bit of custom CSS and add proper styling for the output.

`headless.min.js` is taken from [this projects](https://gitlab.com/mkamner/tech-playground-headless) releases.
I'm happy to take care of updates and related testing in the future.

Looking forward to any feedback you want to share!


mholt:
Very cool! I'm excited to kick this around a bit. (Sorry, been busy with some things lately.) This is on my radar to circle back to soon :)

PR #483, opened 2025-07-01
Corrected file permissions

born2webdesign:
Making sure keyring and repo files are world-readable (o/w error on ` apt update` )

### DIFF
--- a/src/docs/markdown/install.md
+++ b/src/docs/markdown/install.md
@@ -63,6 +63,8 @@ After installing, please read the [service usage instructions](/docs/running#usi
 <pre><code class="cmd"><span class="bash">sudo apt install -y debian-keyring debian-archive-keyring apt-transport-https curl</span>
 <span class="bash">curl -1sLf 'https://dl.cloudsmith.io/public/caddy/stable/gpg.key' | sudo gpg --dearmor -o /usr/share/keyrings/caddy-stable-archive-keyring.gpg</span>
 <span class="bash">curl -1sLf 'https://dl.cloudsmith.io/public/caddy/stable/debian.deb.txt' | sudo tee /etc/apt/sources.list.d/caddy-stable.list</span>
+<span class="bash">chmod o+r /usr/share/keyrings/caddy-stable-archive-keyring.gpg</span>
+<span class="bash">chmod o+r /etc/apt/sources.list.d/caddy-stable.list</span>
 <span class="bash">sudo apt update</span>
 <span class="bash">sudo apt install caddy</span></code></pre>
 
@@ -71,6 +73,8 @@ After installing, please read the [service usage instructions](/docs/running#usi
 <pre><code class="cmd"><span class="bash">sudo apt install -y debian-keyring debian-archive-keyring apt-transport-https curl</span>
 <span class="bash">curl -1sLf 'https://dl.cloudsmith.io/public/caddy/testing/gpg.key' | sudo gpg --dearmor -o /usr/share/keyrings/caddy-testing-archive-keyring.gpg</span>
 <span class="bash">curl -1sLf 'https://dl.cloudsmith.io/public/caddy/testing/debian.deb.txt' | sudo tee /etc/apt/sources.list.d/caddy-testing.list</span>
+<span class="bash">chmod o+r /usr/share/keyrings/caddy-testing-archive-keyring.gpg</span>
+<span class="bash">chmod o+r /etc/apt/sources.list.d/caddy-testing.list</span>
 <span class="bash">sudo apt update</span>
 <span class="bash">sudo apt install caddy</span></code></pre>
 


mholt:
Thanks; what is the issue though, exactly? I believe I've also used the current command sequence myself without problems...

born2webdesign:
Likely not with a umask of 0027 ;) Apparently it has to be read by some user other than root during apt operations (at least on my Debian bookworm system).

src/docs/markdown/install.md
```diff
@@ -63,6 +63,8 @@ After installing, please read the [service usage instructions](/docs/running#usi
 <pre><code class="cmd"><span class="bash">sudo apt install -y debian-keyring debian-archive-keyring apt-transport-https curl</span>
 <span class="bash">curl -1sLf 'https://dl.cloudsmith.io/public/caddy/stable/gpg.key' | sudo gpg --dearmor -o /usr/share/keyrings/caddy-stable-archive-keyring.gpg</span>
 <span class="bash">curl -1sLf 'https://dl.cloudsmith.io/public/caddy/stable/debian.deb.txt' | sudo tee /etc/apt/sources.list.d/caddy-stable.list</span>
+<span class="bash">chmod o+r /usr/share/keyrings/caddy-stable-archive-keyring.gpg</span>
+<span class="bash">chmod o+r /etc/apt/sources.list.d/caddy-stable.list</span>
```
francislavoie, line 67: Are you sure the sources list also needs to be readable, or just the key?

src/docs/markdown/install.md
```diff
@@ -63,6 +63,8 @@ After installing, please read the [service usage instructions](/docs/running#usi
 <pre><code class="cmd"><span class="bash">sudo apt install -y debian-keyring debian-archive-keyring apt-transport-https curl</span>
 <span class="bash">curl -1sLf 'https://dl.cloudsmith.io/public/caddy/stable/gpg.key' | sudo gpg --dearmor -o /usr/share/keyrings/caddy-stable-archive-keyring.gpg</span>
 <span class="bash">curl -1sLf 'https://dl.cloudsmith.io/public/caddy/stable/debian.deb.txt' | sudo tee /etc/apt/sources.list.d/caddy-stable.list</span>
+<span class="bash">chmod o+r /usr/share/keyrings/caddy-stable-archive-keyring.gpg</span>
```
francislavoie, line 66: Can you show the actual error you get without this, with a clean system install? I know it's a tedious ask if you've already made it through this, but I've never been able to replicate a problem although I've heard some people say they had problems.

src/docs/markdown/install.md
born2webdesign: For operation as root, the key file will be enough, if I use an `apt search` as a normal user, however, the repo file needs to be readable as well.

src/docs/markdown/install.md
born2webdesign: `apt update` as root w/o key file being readable:
```
An error occurred during the signature verification. The repository is not updated and the previous index files will be used. GPG error: https://dl.cloudsmith.io/public/caddy/stable/deb/debian any-version InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY ABA1F9B8875A6661
W: Failed to fetch https://dl.cloudsmith.io/public/caddy/stable/deb/debian/dists/any-version/InRelease  The following signatures couldn't be verified because the public key is not available: NO_PUBKEY ABA1F9B8875A6661
W: Some index files failed to download. They have been ignored, or old ones used instead.
```

`apt search` as user w/o repo file being readable:
`W: Unable to read /etc/apt/sources.list.d/caddy-stable.list - open (13: Permission denied)`

As I mentioned before, this is on a system with a umask of `0027` – with `0022` this problem won't occur (as the files would be world-readable). That, to me, still is a ‘clean install’, though.
And it certainly doesn’t hurt to set the permission if already in place; w/o the permission, some people will trip.

francislavoie: Alright. I guess I'm surprised why `0027` would ever be the root umask, seems problematic :man_shrugging:

francislavoie (approved):
Thanks! I find it unfortunate we need this, but at least it should smooth things over for some users.

src/docs/markdown/install.md
born2webdesign: Thanks – I suppose you could also revert this and only add it as a N.B. paragraph and a separate block below the previous commands (that do work for most people)? Whichever you prefer …

born2webdesign: I only set the umask to `0027` in my .bashrc, so files created by the system aren't affected, but manually created ones aren't world-readable, by default. Makes sense to me ;)

mholt: This is also the only time I've heard of that umask. Maybe it's niche?

I would be okay with reverting personally. Still feel weird opening permissions on the files.

francislavoie: @mholt it's the other way around. By default on most systems, the files _are_ created readable. But people with 0027 umask (more restrictive) have the files created _without_ read which is why installation then fails, so they need to chmod the two files (or at least the keyring) before trying to install.

mholt: @francislavoie Ah, okay. Yeah, this does seem unusual then.

born2webdesign: Yea, unusual maybe – but not unreasonable IMO ;)
Also, I forgot to add the `sudo` for the `chmod` commands, so a revert would be best – but I still think it would be a good addition as a separate N.B. paragraph (w/ `sudo chmod`).

PR #485, opened 2025-08-14
Update copr installation instructions

carlwgeorge:
Since these instructions were originally written, a few things have changed.

* Fedora has switched from dnf to dnf5
* CentOS/RHEL 7 are now EOL and no longer have packages in the copr

This refreshes the instructions to reflect the current state.


francislavoie:
Thanks!

PR #488, opened 2025-08-19
build: add Docker flags to cache builds and deps

mohammed90:


steffenbusch:
I was not aware of that feature, which works with podman too. I have tested it today and it’s a huge time saver for me on recurring xcaddy build runs with minor changes on referenced plugins and their versions.
Thank you very much!

PR #491, opened 2025-08-27
docs: Add `.test` as an internal-only hostname

jaydrogers:
## ⚠️ This is PR depends on a change in "certmagic"
**Do not merge this PR until the related PR below has been merged & released.** This change will only make sense if this change in `certmagic` is made available:

- https://github.com/caddyserver/certmagic/pull/351

## What this PR does
- Adds `.test` as an "internal-only host name"

## More information on this change
See the `certmagic` PR above for more information explaining this proposed change.

### DIFF
--- a/src/docs/markdown/automatic-https.md
+++ b/src/docs/markdown/automatic-https.md
@@ -117,7 +117,7 @@ All hostnames (domain names) qualify for fully-managed certificates if they:
 
 In addition, hostnames qualify for publicly-trusted certificates if they:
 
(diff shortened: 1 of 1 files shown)


jaydrogers:
After [further discussion](https://github.com/caddyserver/certmagic/pull/351#issuecomment-3267168720), this PR should not be merged.

It is not safe to assume that `.test` cannot solve ACME challenges. It is possible for ACME challenges to be completed internally, there for it could create a trusted certificate.

PR #493, opened 2025-09-06
add version notice for excluded hidden files for import

JL710:
This PR intents to warn users that use an older version of caddy, like on debian where the version is currently on 2.6.2, that they must be careful with hidden files (i.e. from editors).

### DIFF
--- a/src/docs/markdown/caddyfile/directives/import.md
+++ b/src/docs/markdown/caddyfile/directives/import.md
@@ -42,7 +42,7 @@ import <pattern> [<args...>] [{block}]
 
 ## Examples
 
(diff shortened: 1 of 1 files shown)


francislavoie:
We don't support versions that old. Nobody should be using 2.6.2 at all. Debian shipped that old version of Caddy without consulting us and they never spent any effort upgrading to a properly new version. If you insist on using it instead of our own apt repo, then please contact them to get them to upgrade Caddy.

JL710:
Oh I missed that. I just installed with the package manager. I switched to the apt repo and should have read the install page first. Sorry for bothering.

PR #496, opened 2025-09-21
Add span attributes to tracing documentation

felix-hilden:
Related to caddyserver/caddy#7269, I'm adding the corresponding documentation for setting custom span attributes. This is still work in progress, but the documentation at least specifies the desired functionality of the associated PR.

### DIFF
--- a/src/docs/markdown/caddyfile/directives/tracing.md
+++ b/src/docs/markdown/caddyfile/directives/tracing.md
@@ -18,11 +18,11 @@ The trace ID and span ID are added to [access logs](/docs/caddyfile/directives/l
 
 ```caddy-d
 tracing {
-	[span <span_name>]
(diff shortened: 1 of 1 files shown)


mholt (approved):
Thanks -- I think this is good. Although I just noticed that enclosing subdirectives entirely in `[ ]` is a little odd, since I think by default we assume that subdirectives, or lines within a directive block, are optional already.

Of course that's not new with your PR, I'm looking at line 21 as well...

No need to block this PR on that, but just something I noticed that might be worth discussing for consistency with the rest of our docs. I'm inclined to remove those `[ ]`, maybe in another commit.

Thank you!

felix-hilden:
Thanks! A small trouble to improve, no worries. The square brackets are now removed!

PR #497, opened 2025-09-24
Add warning for keepalive duration and connection resets

### LINKED ISSUE #6452: Reverse proxy upstream keepalive
Hello, thanks for Caddy, it's a pleasure to use it for its easy config and its "it just works" approach.

### Issue

I've spent hours trying to understand why Caddy was randomly returning 502s for some proxied calls to my Node.js backend. It was logging the following message which made me pointlessly look for issues in my network stack:

    read tcp #.#.#.#:#->#.#.#.#:#: read: connection reset by peer

### Resolution

I have finally identified the root cause which is that the default http transport keepalive duration setting in Caddy (2 minutes) widely exceeds the default Node.js HTTP server keepAliveTimeout setting (5 seconds).

### Improvements

I believe it would be worth doing the following:
1. making Caddy reopen a connection instead of returning a 502 to the client (turning the problem into a warning instead of an error)
2. pointing out in the documentation that Caddy enables a keepalive by default (should that be revisited?)
3. pointing out in the documentation that the keepalive duration must be below the upstream server's own keepalive timeout
4. improving the log in case the connection that was kept alive happens to be closed

### Rationale

This issue is not specific to Caddy and you can find many articles online of people spending lots of time on similar issues for each proxy out there that has a long keepalive enabled by default. So it would fit Caddy's philosophy to provide something that works better out-of-the-box.

This issue tends to appear for small apps that suddenly have more users. Backends that get little activity hit Caddy's timeout while the ones with a lot of activity don't hit Node's timeout. But for a small server which sees a few bursts of activity, suddenly a lot of unexplainable 502s pop up as many calls fall in the 5sec < x < 2min gap.

joanrieu:
Added a warning about potential connection reset errors with HTTP/1.1 upstreams when keepalive duration exceeds the upstream server's timeout.

Resolves https://github.com/caddyserver/caddy/issues/6452

### DIFF
--- a/src/docs/markdown/caddyfile/directives/reverse_proxy.md
+++ b/src/docs/markdown/caddyfile/directives/reverse_proxy.md
@@ -604,6 +604,8 @@ transport http {
 
 - **keepalive** <span id="keepalive"/> is either `off` or a [duration value](/docs/conventions#durations) that specifies how long to keep connections open (timeout). Default: `2m`.
 
+  ⚠️ Requests to HTTP/1.1 upstreams may fail due to "connection reset by peer" errors if the keepalive duration exceeds the upstream server's keepalive timeout. Idempotent requests will be retried by Go's HTTP transport, but Caddy will respond with status code 502 in other cases.
+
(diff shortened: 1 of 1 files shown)


mholt (approved):
Thanks!

PR #500, opened 2025-10-07
Remove dud restriction for `caddy reload`

matburton:
I've given it a quick test and you can specify `caddy reload --address unix//run/api.socket` or similar without issue.


francislavoie:
Thanks for confirming!

PR #512, opened 2025-12-27
Proposal: Update Ansible community role

sebdanielsson:
Updates the community Ansible role for installing Caddy.

What the new role offers:

-  Build Caddy with plugins using xcaddy.
-  Build Caddy for almost any distro for amd64 and arm64 systems (Tests for Debian-based, Fedora-based, Arch-based).
-  Customize values like GOPROXY (Useful for enterprises that might want to use their own Go proxy).

### DIFF
--- a/src/docs/markdown/install.md
+++ b/src/docs/markdown/install.md
@@ -181,11 +181,19 @@ You may need to adjust the Windows firewall rules to allow non-localhost incomin
 
 ## Ansible
 
-_Note: This is a community-maintained installation method._
+_Note: These are community-maintained installation methods._
+
+xcaddy - Build Caddy with plugins:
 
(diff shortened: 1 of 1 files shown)


francislavoie:
FYI @nvjacobo in case you want to give your input

sebdanielsson:
@francislavoie what do you think?☺️

francislavoie:
I think I'd rather we just list both rather than removing the other one. I think it would be rude to @nvjacobo to do so without getting their feedback if they still plan on maintaining theirs.

sebdanielsson:
> I think I'd rather we just list both rather than removing the other one. I think it would be rude to @nvjacobo to do so without getting their feedback if they still plan on maintaining theirs.

Ah, sorry about that. I've added the current role back. I assumed it was a stale project because it hasn't been updated with official support for current distro versions.

PR #515, opened 2026-01-03
Fix mixup in route docs

Maxr1998:
The word "second" was used twice, and the directives visible in the sample below were swapped.
I feel that even with my fix the section is still a bit confusing, since the paragraph above describes a different order. Since the line order doesn't make a difference in the first place, one could also omit the "first" and "second" and just write this instead:

> However, there may be occasions where the `file_server` has a more specific matcher than `redir`. In other words, you want to redirect in the general case, and serve only a specific file.


francislavoie (approved):
Wow, yeah that sentence is very backwards 😂 Thanks!

PR #517, opened 2026-01-16
Address another source of subdomain leaks

Kioubit:


mholt (approved):
"Ironic. DNSSEC could save others from security leaks, but not itself." -Palpatine (probably)

Thanks!

PR #519, opened 2026-01-20
Add 88x31 button for Caddy

mat-1:
I made a Caddy 88x31 button a while ago for my website (which has now spread to [several others](https://matdoes.dev/buttons#af60dfae333038fc162db5d39868229b)), and since Caddy already has a few 88x31s at the bottom of its website, I figure this one would fit in nicely. :3

Here's what it looks like next to the other current ones:
image

If it matters, I license the button image as CC0. Also, I already losslessly compressed it with the max level on Oxipng.


mholt:
That's a fun idea! Let me think about this for a moment. Is it possible to have it match the size of the other buttons a little more closely? It looks a bit small. Just by a few pixels.

mat-1:
I'm glad you like it!

I sized it as exactly 88x31 pixels since that's the classic and by far most common resolution for these types of buttons. The Internet Explorer button on Caddy's website is also 88x31, but the rest are different (Firefox one is 110x32, Netscape is 90x30, W3C one is 80x30, and the hit counter is 96x26).

I suppose the easiest solution would be to move it to be around the similarly-sized images so it doesn't look as out of place:
image

Alternatively, the Firefox one could be replaced with an image that is actually 88x31 (the image I chose here is also used on [a bunch of random personal websites](https://matdoes.dev/buttons#d3759805fe0ac49834e09e43e6348438)):
image

What do you think?

mholt:
I always thought that Firefox button looked a little weird :)

Now that you mention it, maybe the Caddy button does look better over by the HTML/CSS buttons. Can we just move it there and then we'll call it a day?

Thanks for the fun contribution!

mat-1:
Sounds good! Moved it to be between the Netscape and HTML/CSS buttons. :)

PR #523, opened 2026-02-12
Simplify installation instructions by removing chmod commands

Justinzobel:
These commands aren't needed.


francislavoie:
They are needed in some situations, actually. https://github.com/caddyserver/website/pull/483

Justinzobel:
Given it isn't needed for most installs, I don't see the need to have it, but they're your docs.

francislavoie:
Given it's needed for _some_ installs, it _must_ be in the steps, otherwise we get complaints from people who fail to install (then get sad, pissed, throw caddy in the bin at the first roadblock).

It costs us nothing to keep it in the steps when it smooths out the process for people who have the issue.

PR #525, opened 2026-02-12
Change setPageTitle to only run on /docs/json pages

### LINKED ISSUE #524: HTML <title> from Markdown file gets overwritten by `setPageTitle` in json-docs.js
Most/all of the Markdown pages have the HTML `<title>` tag populated from the Markdown title. However, then `json-docs.js` gets loaded and unconditionally overwrites the title with logic that seems to be for `/docs/json` requests only. This leads to issues particularly with nested pages like [/docs/quick-starts/static-files](https://caddyserver.com/docs/quick-starts/static-files), where `<title>` gets changed to 

> -starts/static-files - Static files quick-start — Caddy Documentation

It looks like just moving the call to `setPageTitle` a couple lines down so it's only called for `/docs/json` request fixes the issue, see screenshot below.

Image

joeldemars:
Fixes [#524](https://github.com/caddyserver/website/issues/524)

### DIFF
--- a/src/resources/js/json-docs.js
+++ b/src/resources/js/json-docs.js
@@ -3,10 +3,10 @@ const jsonDocsPathPrefix = "/docs/json/";
 var configPath = window.location.pathname.slice(jsonDocsPathPrefix.length);
 var pathComponents = configPath.split('/');
 
-setPageTitle();
-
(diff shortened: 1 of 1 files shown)


francislavoie (approved):
Thanks! This was a regression from the redesign we pushed out a week or two ago.

PR #526, opened 2026-02-12
Language clarification/typo

MarcelWaldvogel:
* Fixed typo "mather"
* For me, "both X and Y are redundant" means that you can **delete both** and it still works. Changed that to "having both X and Y is redundant", i.e., one is enough. (I am not a native speaker, though.)


francislavoie (approved):
Thanks!

I was being terse in concepts.md since I wanted to avoid the text wrapping in the table for visual consistency, but looks like other things in the same column were already slightly longer, so adding `having` shouldn't cause any wrapping.

PR #530, opened 2026-03-08
docs: add sudo prefix to chmod commands in Debian/Ubuntu install instructions

ihanikos:
The chmod commands modify files in /usr/share/keyrings/ and /etc/apt/sources.list.d/ which require root permissions.

Related: #523 (closed) - chmod commands are needed, but they need sudo prefix.


francislavoie (approved):
Thanks! Silly oversight.

PR #533, opened 2026-03-28
Add verified install script

ropoctl:
POSIX sh installer that downloads Caddy exclusively from official GitHub Releases, verifies SHA512 checksums, and optionally verifies Cosign/Sigstore signatures.

Served at caddyserver.com/install.sh

Usage: curl -fsSL https://caddyserver.com/install.sh | sh

No third-party domains contacted — all downloads from github.com.

### DIFF
--- a/src/install.sh
+++ b/src/install.sh
@@ -1,0 +1,228 @@
+#!/bin/sh
+# Secure Caddy Installer
+# Downloads Caddy from official GitHub Releases (caddyserver/caddy)
+# Verifies SHA512 checksums and optionally Cosign signatures
+#
+# Usage:
+#   curl -fsSL https://caddyserver.com/install.sh | sh
+#   curl -fsSL https://caddyserver.com/install.sh | sh -s -- --dry-run
+#   curl -fsSL https://caddyserver.com/install.sh | sh -s -- --version v2.11.2
+#
+# All downloads come exclusively from github.com/caddyserver/caddy
+# No third-party domains are contacted.
+
+set -eu
+
+GITHUB_REPO="caddyserver/caddy"
+GITHUB_API="https://api.github.com/repos/${GITHUB_REPO}"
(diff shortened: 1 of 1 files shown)


francislavoie:
Sorry, we won't do this. We don't want to encourage people to curl-bash, it's very poor security hygiene. It's much better to just follow instructions on https://caddyserver.com/docs/install for your platform, especially because this wires up automated package updates as well.

ropoctl:
I wouldn't condone it either, but if you don't someone else already has
image
https://x.com/mholt6/status/2037588986089349511

coolaj86:
Goreleaser builds install scripts for GitHub, which are what many projects use.

The modern arguments against web-based installers are religious, not technical - meaning when you ask someone who says you _shouldn't_ do it _why_, they typically aren't old enough to know the context of the advice - just that their sensei said it (rightfully), so they repeat it (mistakenly).

To summarize:
- don't install from untrusted sources: ✅correct \
  I trust caddy to provide correct sources more than apt (or Webi) or anywhere else
- don't install from unverified sources: ✅ correct \
  aes-gcm over https provides both tamper proofing and proof of providence
- don't run potentially truncated commands: ✅ correct \
  all modern installers now use command groups (or equivalent) to ensure that a script is fully parsed (syntactically correct) before the first line runs

The wisdom of the ancients was sound and correct - they understood:
- the fickle nature of floppy disks
- the absurd insecurity of plain-text downloads
- the necessity of ensuring that when your modem cuts off in the middle of transmission you don't `rm -rf /` yourself on a packet boundary
- and to not download (or take floppies) from _untrusted_ sources - the trolls of the forums and message boards

But even many children of the 90s don't actually know what the ftp their predecessors were warning them against.

And you can't argue with them - because it's not a technical argument - it's an argument of love and respect for those who paved the way before with wisdom and strength. ❤️ 💪

PR #536, opened 2026-05-01
docs: Document OTLP metrics export

steadytao:
## Summary

Documents Caddy's OTLP metrics export support added by caddyserver/caddy#7664.

Updates:
- the monitoring guide to cover both Prometheus scraping and OTLP push
- the global `metrics` option docs to include `otlp`
- the `metrics` directive cross-reference text
- quick file cleanup run on each of the files touched *(trailing sapces)*

## Details

Change covered in caddyserver/caddy#7369, brought in by caddyserver/caddy#7664. I have confirmed that PR satisfies OLTP standards *(https://pkg.go.dev/go.opentelemetry.io/contrib/exporters/autoexport)*. 

`metrics { otlp }` exports the same metrics gathered for the Prometheus endpoint to an OpenTelemetry Protocol endpoint.

The exporter is configured through standard OpenTelemetry `OTEL_*` environment variables, including endpoint, protocol, headers, export interval and exporter selection.

### DIFF
--- a/src/docs/markdown/metrics.md
+++ b/src/docs/markdown/metrics.md
@@ -1,8 +1,8 @@
 ---
-title: Monitoring Caddy with Prometheus metrics
+title: Monitoring Caddy with metrics
 ---
 
-# Monitoring Caddy with Prometheus metrics
+# Monitoring Caddy with metrics
 
 Whether you're running thousands of Caddy instances in the cloud, or a single
 Caddy server on an embedded device, it's likely that at some point you'll want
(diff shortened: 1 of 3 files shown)


src/docs/markdown/metrics.md
```diff
@@ -77,6 +77,47 @@ You can then start up Prometheus like this:
...
+This is useful for OTLP-native observability stacks, such as an OpenTelemetry
+Collector, Grafana Alloy, Honeycomb, or other systems that receive OTLP metrics
+directly.
+
+Enable OTLP metrics export with the `otlp` option:
+
+```caddy
+{
+	metrics {
+		otlp
```
mholt, line 92: Before we document this, I wonder if enabling OTLP should be as simple as `metrics otlp` (no block). Any downsides to that?

steadytao: I wouldn't say there are particularly *downsides* to it. More so that it matches `metrics` as the global option with OTLP as a subdirective and it leaves room for possible future metrics subdirectives without making the top-level form ambiguous.

francislavoie: I see no reason to add shortcuts like that eagerly. To be clear, this isn't the first sub-option to metrics, we already have `per_host` etc. I don't see why `oltp` would get special care in comparison.

mholt: Alrighty. Good enough for me!

mholt (approved):

francislavoie:
We shouldn't have merged @mholt this is docs for the next release

mholt:
Yeah, but that will be soon, and unlikely to really update the website before then anyway.

PR #537, opened 2026-05-04
Fix nav.html include path in index.html

stupidloud:
Fix a broken include path in `src/docs/json/index.html` by changing `{{include "/includes/docs/nav.html"}}` to `{{include "/old/includes/docs/nav.html"}}`, so the JSON docs page loads the correct nav partial.

### DIFF
--- a/src/docs/json/index.html
+++ b/src/docs/json/index.html
@@ -12,7 +12,7 @@
 	<body>
 		{{include "/includes/header.html" ""}}
 		<main>
-			{{include "/includes/docs/nav.html"}}
+			{{include "/old/includes/docs/nav.html"}}
 			<div class="article-container">
 				<div class="paper" id="paper1"></div>
 				<div class="paper" id="paper2"></div>


mholt:
What? Why? What is broken?

stupidloud:
Hi @mholt, sorry for the terse PR description — let me explain what I ran into.

On the current `master`, visiting `/docs/json/` fails to render the docs nav because `src/docs/json/index.html` includes `/includes/docs/nav.html`, but that file doesn't exist in the repo. `src/includes/docs/` only contains `details.html`, `hovercard.html`, and `renderbox.html`. The actual `nav.html` lives at `src/old/includes/docs/nav.html`.

The sibling page `src/docs/modules/index.html` already includes the nav from `/old/includes/docs/nav.html`, so this PR just brings `json/index.html` in line with it. Happy to take a different approach if you'd prefer the file moved/symlinked instead — just let me know.

PR #545, opened 2026-06-13
Document that file_server's hide can be used multiple times

luccinmasirika:
Companion docs for caddyserver/caddy#7817 (closes the `needs docs` note there), which makes repeated `hide` subdirectives append instead of overwrite.

This adds one sentence to the `hide` description noting that multiple uses combine rather than replacing each other, which is what lets imported snippets contribute common hides alongside site-specific ones. No other changes. Related issue: caddyserver/caddy#7815.

Assistance disclosure: I used an AI assistant to help locate the right spot and draft the wording; I reviewed it and confirmed it matches the behavior in the linked PR.


src/docs/markdown/caddyfile/directives/file_server.md
```diff
@@ -60,7 +60,7 @@ file_server [<matcher>] [browse] {
...
+- **hide** <span id="hide"/> is a list of files or folders to hide; if requested, the file server will pretend they do not exist. Accepts placeholders and glob patterns. Note that these are _file system_ paths, NOT request paths. In other words, relative paths use the current working directory as a base, NOT the site root; and all paths are transformed to their absolute form before comparisons (if possible). Specifying a file name or pattern without a path separator will hide all files with a matching name regardless of its location; otherwise, a path prefix match will be attempted, and then a globular match. Since this is a Caddyfile config, the active configuration file(s) will be added by default. The subdirective may be specified more than once; each use appends to the list rather than replacing it, which lets an imported snippet contribute common hides that combine with site-specific ones. Hide comparisons are case-sensitive; on case-insensitive filesystems, a differently-cased request path may still resolve to the same on-disk path, so `hide` should not be treated as a security boundary for sensitive paths.
```
steadytao:
```suggestion
- **hide** <span id="hide"/> is a list of files or folders to hide; if requested, the file server will pretend they do not exist. Accepts placeholders and glob patterns. Note that these are _file system_ paths, NOT request paths. In other words, relative paths use the current working directory as a base, NOT the site root; and all paths are transformed to their absolute form before comparisons (if possible). Specifying a file name or pattern without a path separator will hide all files with a matching name regardless of its location; otherwise, a path prefix match will be attempted, and then a globular match. Since this is a Caddyfile config, the active configuration file(s) will be added by default. The subdirective may be specified more than once; each use appends to the list rather than replacing it, which lets an imported snippet contribute common hides that combine with site-specific hides. Hide comparisons are case-sensitive; on case-insensitive filesystems, a differently-cased request path may still resolve to the same on-disk path, so `hide` should not be treated as a security boundary for sensitive paths.
```

steadytao (changes requested):

steadytao (approved):
LGTM. Merge for v2.11.5

PR #548, opened 2026-06-19
docs: Recommend docker exec to reload config

timhallmann:
`docker compose kill` marks the container as manually stopped. A restart of dockerd will thus not restart caddy when running with the recommended `restart: unless-stopped`.

This problem can be avoided by issuing kill within the container.


francislavoie:
It's not correct to remove `compose` because that's what does container mapping by service name. If you used docker compose then the container name will not be simply `docker`.

This seems like a bug with Docker Compose that should be reported upstream. It's nonsense that it doesn't exclude USR1 from counting as stopping the container.

timhallmann:
Right, you're not setting `container_name: caddy`, which I do... Missed that, sorry.

Until/If docker composes fixes that, should the docs simply mention this as a caveat?

PR #550, opened 2026-06-24
Install: Use long form easy to read arguments in apt

Justinzobel:


mholt:
(Why? Was the original hard to read/understand? I feel like -y in apt is pretty common. Like Debian 101)

Justinzobel:
> (Why? Was the original hard to read/understand? I feel like -y in apt is pretty common. Like Debian 101)

For all long time users of course, but in documentation it is always best to aim for the newest user of any software.

mokshendraisglobal (approved):

PR #556, opened 2026-07-10
Document -1 as the value that disables roll_keep

### LINKED ISSUE #7863: Setting roll_keep to 0 does not disable it
### Issue Details

Documentation about the `roll_keep` log rotation option at https://caddyserver.com/docs/caddyfile/directives/log#roll_keep states the following:

> roll_keep is how many log files to keep before deleting the oldest ones. Triggers when a new log file is created.
> To disable roll_keep, set the value to 0.

But setting the value to 0 in Caddyfile seems to have no effect and keeps a maximum of 10 log files instead. Example Caddyfile to reproduce the issue:

```
mywebsite:80 {
    log {
        output file /var/log/caddy/mywebsite.log {
            roll_interval 1s
            roll_keep 0
            roll_keep_for 1d
        }
        format json
    }
    respond "OK"
}
```

```bash
# caddy --version
v2.11.4 h1:XKxkMTgNSizEvKG6QHue6cAsFOteU2qA61w2tKkCWi0=
# ls -la /var/log/caddy/ | grep log.gz | wc -l
10
```

This could be caused by this: https://github.com/caddyserver/caddy/blob/4e62095245005b17f290dd1e4d9afe983df180ee/modules/logging/filewriter.go#L258-L260

Setting the value to -1 instead removes the limit.

### Assistance Disclosure

AI not used

### If AI was used, describe the extent to which it was used.

_No response_

Socialpranker:
Fixes caddyserver/caddy#7863

## Summary

The `roll_keep` docs say to set the value to `0` to disable the retention limit, but that's not what the code does: `modules/logging/filewriter.go` treats a `RollKeep` of `0` as "not set" and falls back to the default of `10` ([filewriter.go#L258-L260](https://github.com/caddyserver/caddy/blob/master/modules/logging/filewriter.go#L258-L260)). The value that actually removes the limit is `-1`, which the underlying rotator (lumberjack) treats as "no limit" for any `MaxBackups <= 0`.

A maintainer confirmed this in the issue thread and noted the fix belongs here (in the docs), not in a behavior change to `caddy` itself.

## Change

One-line correction to the `roll_keep` description: documents `-1` as the disabling value and explains that `0` falls back to the default of `10` instead of disabling anything.

## Test plan

- Read the corrected sentence against `filewriter.go`'s actual `RollKeep`/`MaxBackups` handling to confirm the new wording matches behavior.
- This is a documentation-only change with no code or build output to run.


Socialpranker:
Friendly ping — this is a one-line docs fix that's been open since 10 July.

The `roll_keep` docs say `0` disables the retention limit, but `filewriter.go` treats `0` as "not set" and falls back to the default of 10; `-1` is the value that actually disables it (caddyserver/caddy#7863, still open).

Should be a quick one whenever someone has a moment.

Socialpranker:
Still here a few weeks on — it's a one-liner, so happy to tweak the wording if the phrasing isn't right, just say so.

Socialpranker:
Heads up for whoever gets to this: #566 opened a duplicate fix for the same issue (caddyserver/caddy#7863) on 8/15. Functionally near-identical (both correct the disable value to `-1`), but this PR's wording also clarifies that `0` isn't a no-op — it falls back to the default of `10` — which #566's doesn't mention. Leaving it to you to pick whichever you'd rather merge; happy to close this one if you'd prefer to take #566 instead.

PR #560, opened 2026-08-09
docs: note admin unix socket directory permissions

### LINKED ISSUE #420: Following the docs leads to service failure
The docs at https://caddyserver.com/docs/caddyfile/options#admin describe how to use a unix socket in order to make the api listen to the unix socket.

```
{
	admin unix//run/caddy-admin.sock
}
``` 

Following this leads to a service failure on restart:

> Oct 01 06:07:00 server1 caddy[56725]: Error: loading initial config: loading new config: starting caddy administration endpoint: permission denied

This happens, if caddy is installed via package manager and running as user caddy. 

But even if I create that socket using:

```
touch /run/caddy-admin.sock
chown caddy: /run/caddy-admin.sock
chmod 0600 /run/caddy-admin.sock
````

Leads to the very same error:

> Oct 01 06:10:14 server1 caddy[56747]: Error: loading initial config: loading new config: starting caddy administration endpoint: permission denied

### PR DESCRIPTION
**Pull request by locker95**
The admin unix socket example uses `/run/caddy-admin.sock`, which fails with permission denied when Caddy runs as the package `caddy` user. Added a short note that the process needs to create the socket path (and that `touch` alone is not enough).

Fixes #420

### CODE REVIEW

File: src/docs/markdown/caddyfile/options.md [hunk 1]
```diff
@@ -330,6 +330,8 @@ To use a [unix socket](/docs/conventions#network-addresses) for the admin API, a
 }
 ```
 
+The process user must be able to create (and replace) the socket path. Packaged installs often run as a non-root user such as `caddy`, which typically cannot write under `/run` itself—use a directory that user owns (for example under the service data dir or a unit `RuntimeDirectory`), or adjust permissions. Creating an empty file with `touch` first is not enough; Caddy needs to bind the socket.
```
```suggestion
The process user must be able to create (and replace) the socket path. Packaged installs often run as a non-root user such as `caddy`, which typically cannot write under `/run` itself. Instead, use a directory that user owns (for example under the service data dir or a unit `RuntimeDirectory`), or adjust permissions. Creating an empty file with `touch` first is not enough; Caddy needs to bind the socket.
```

PR #569, opened 2026-09-09
docs: document validate conflicts with exclusive resources

faiyazrahmann:
Docs follow up for caddyserver/caddy#8004.

`caddy validate` provisions modules for real, so it fails when a running
instance already holds an exclusive resource. `acme_server` is the common
case: its database can only be open in one process at a time, so validate
times out even though the config is fine.

Adds an explanation in the `caddy validate` section, and a short note on the
`acme_server` directive pointing back to it. Both say to validate before
starting Caddy, or use `caddy reload` to check against a running instance.

@steadytao you offered a review on this, so tagging you here.


src/docs/markdown/command-line.md
```diff
@@ -688,6 +688,8 @@ Similarly to `caddy upgrade`, replaces the current Caddy binary with the latest
 
 Validates a configuration file, then exits. This command deserializes the config, then loads and provisions all of its modules as if to start the config, but the config is not actually started. This exposes errors in a configuration that arise during loading or provisioning phases and is a stronger error check than merely serializing a config as JSON.
 
+Because modules are provisioned for real, this command is best run before starting Caddy. Some modules acquire exclusive resources while being provisioned, and a second Caddy process cannot acquire them while a running instance holds them. For example, the [`acme_server`](/docs/caddyfile/directives/acme_server) directive opens a database that only one process can have open at a time, so validating a config that uses it while Caddy is already running fails with a database timeout, even though the config itself is fine. To check a config against a running instance, use [`caddy reload`](#caddy-reload) instead: it loads the new config in the running process, and the previous config keeps running if the new one fails.
```
steadytao: Should say that it provisions within the running process, retains the active configuration on failure and applies the new configuration on success. Make the same distinction on the acme_server page.

steadytao (changes requested):

steadytao (approved):
LGTM.

PR #570, opened 2026-09-16
Document the `tls_automate_names` global option

IslamElsayed:
Documents the `tls_automate_names` global option added in caddyserver/caddy#8015, as @steadytao asked for on that PR.

The name is the one @francislavoie settled on there, so this should not need renaming — but if it does move, it is this one entry plus the line in the syntax summary.

**One thing I could not determine and would like corrected rather than guessed at:** I wrote *"(Requires Caddy 2.12 or newer.)"* on the basis that v2.11.4 is the latest release, so the next minor would be 2.12. If the option lands in a different version, that line is wrong.

I checked both code samples against a build of the caddy PR rather than writing them from memory:

- the `tls_automate_names` example adapts to `certificates.automate: ["*.example.com"]` with a route only for `foo.example.com`
- the empty-site-block example does produce the `*.example.com` route the text warns about

The other claims in the entry — that a name with its own site block keeps that site's settings, that the option accumulates when repeated, and that a Caddyfile of only global options is valid with it set — are each covered by an adapt test in the code PR.

Disclosure: I use AI assistance when working on patches. I wrote and checked this entry myself, and verified every example and claim above against a build.


src/docs/markdown/caddyfile/options.md
```diff
@@ -474,6 +475,33 @@ This means that if you wish to serve your site over HTTP, you should change your
 ```
 
 
+##### `tls_automate_names`
+Manages certificates for the given names without serving them. The names get the same certificate management [Automatic HTTPS](/docs/automatic-https) gives the names in your site blocks, but no route is added, so Caddy does not respond for them.
```
steadytao: State that this is an explicit request and still enables certificate management when `auto_https off` or `disable_certs` is configured. The implementation deliberately behaves like `force_automate`; describing it only as the same management provided by Automatic HTTPS does not make that precedence clear.

src/docs/markdown/caddyfile/options.md
```diff
@@ -474,6 +475,33 @@ This means that if you wish to serve your site over HTTP, you should change your
...
+
+Use this for a name you need a certificate for but do not serve with Caddy's HTTP server: a wildcard that only covers other sites, a mail server, or a name handled by a [layer 4](https://github.com/mholt/caddy-l4) app. A Caddyfile containing only global options is valid when this option is set.
+
+A name that also has its own site block keeps that site's certificate settings. May be repeated; the names accumulate.
+
+(Requires Caddy 2.12 or newer.)
```
steadytao: Remove this version claim until the target release is confirmed.

steadytao (changes requested):
Otherwise looks alright.

IslamElsayed:
Updated the version note from "2.12 or newer" to "2.11.5 or newer" in 30a0dc0, following @francislavoie and @steadytao settling on 2.11.5 for this in caddyserver/caddy#8015 — it was the one thing in here I had flagged as a guess rather than a fact.

steadytao:
Cheers. Just the remaining comments then should be good.

IslamElsayed:
Both addressed — and apologies for the delay on the first one, I missed your inline comments for four days because I was filtering reviews by date and they predated the window I was looking at. Entirely my fault.

**Precedence (line 479)** — rewritten in 712f690:

> Listing a name here is an explicit request, so it takes precedence over the general switch: certificates are still managed for it when `auto_https` is set to `off` or `disable_certs`. In that it behaves like the `tls` directive's `force_automate`, which forces automation for a site even when other managed certificates apply.

I checked both halves rather than just the one I had a test for — adapting `{ auto_https off; tls_automate_names mail.example.com }` and the same with `disable_certs` each leave `certificates.automate: ["mail.example.com"]` intact. caddyserver/caddy#8015 pins the `off` case in `tls_automate_names_auto_https_off.caddyfiletest`.

**Version claim (line 485)** — you asked for it removed until the target was confirmed. It has been since, in caddyserver/caddy#8015, so it now reads "2.11.5 or newer" rather than the 2.12 guess. Say the word if you would still rather it came out entirely until the release actually ships.

steadytao (approved):

PR #571, opened 2026-09-16
fix: trusted_proxies_strict: Downstream proxies, NOT upstream proxies

marco-m:
The deployment the text is referring to:

       Internet | Internal
                |
    client --- edge-proxy --- Caddy --- upstream

The text was calling the edge-proxy "upstream proxy", while it is, from the point of view of Caddy, a "downstream proxy".

When thinking about upstream/downstream I always have to pause, so I think it is common to get confused :-)


francislavoie:
Thanks, yeah I agree it is a mental hurdle :sweat_smile:

